PR #2, opened 2016-03-03
[WIP] Fix EZP-25536: eZSupportTools review

glye:
> Fixes https://jira.ez.no/browse/EZP-25536
> Status: Work in progress

The ez-support-tools repo was created without peer review, this aims to fix that. There are three existing commits in master, of these only the first is really interesting, the others are trivial.
1. https://github.com/ezsystems/ez-support-tools/commit/116a6d7290a316183e672762944d7c41cf84cd5c
2. https://github.com/ezsystems/ez-support-tools/commit/0e0ea114cf3d6607525eb9339b20627e477b4e90
3. https://github.com/ezsystems/ez-support-tools/commit/cf61fe9a2fe40ac9d6d3150ac225e6b39806d691

My TODO:
- [ ] Find a better way to inject the InfoProviders? (using a variadic constructor at the moment, not sure if legit)
- [ ] Get the platformui integration working, so existing sysinfo code in platformui can be removed/deprecated
- [x] ~~Command, see https://github.com/ezsystems/ez-support-tools/pull/1~~
- [ ] Tests

Feedback from @dpobel:
- [x] D1) Let InfoProvider return a ValueObject instead of array https://github.com/ezsystems/ez-support-tools/commit/116a6d7290a316183e672762944d7c41cf84cd5c#commitcomment-16469472
- [ ] D2) Separate template from InfoProvider https://github.com/ezsystems/ez-support-tools/commit/116a6d7290a316183e672762944d7c41cf84cd5c#commitcomment-16468867
- [x] D3) Add `ezcSystemInfo` dependency to composer.json https://github.com/ezsystems/ez-support-tools/commit/116a6d7290a316183e672762944d7c41cf84cd5c#commitcomment-16468900
- [x] D4) Change YUI prefix `ezconf` to `ezst` https://github.com/ezsystems/ez-support-tools/commit/116a6d7290a316183e672762944d7c41cf84cd5c#commitcomment-16469054

Feedback from @bdunogier:
- [ ] B1) Make routing paths less generic. (Maybe: `ezst/sysinfo/*`?) https://github.com/ezsystems/ez-support-tools/commit/116a6d7290a316183e672762944d7c41cf84cd5c#commitcomment-16494288

### DIFF
--- a/Controller/EzSupportToolsController.php
+++ b/Controller/EzSupportToolsController.php
@@ -1,0 +1,64 @@
+<?php
+
+/**
+ * File containing the EzSupportToolsController class.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzSupportTools\Controller;
+
+use eZ\Publish\Core\MVC\Symfony\Security\Authorization\Attribute;
+use EzSystems\PlatformUIBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\Response;
+
+class EzSupportToolsController extends Controller
+{
+    /**
+     * @var \EzSystems\EzSupportTools\InfoProvider\InfoProviderInterface[] Info providers
+     */
+    protected $infoProviders;
+
+    public function __construct(...$infoProviders) // TODO pretty sure this isn't the way to do this
+    {
+        $this->infoProviders = $infoProviders;
+    }
+
+    public function performAccessChecks()
+    {
+        parent::performAccessChecks();
+        $this->denyAccessUnlessGranted(new Attribute('setup', 'system_info'));
+    }
+
+    /**
+     * Renders the system information page.
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function infoAction()
+    {
+        $infoArray = ['infoProviders' => []];
+        foreach ($this->infoProviders as $infoProvider) {
+            $infoArray['infoProviders'][$infoProvider->getIdentifier()] = [
+                'template' => $infoProvider->getTemplate(),
+                'info' => $infoProvider->getInfo(),
+            ];
+        }
+
+        return $this->render('eZSupportTools:SystemInfo:info.html.twig', $infoArray);
+    }
+
+    /**
+     * Renders a PHP info page.
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function phpinfoAction()
+    {
+        ob_start();
+        phpinfo();
+        $response = new Response(ob_get_clean());
+
+        return $response;
+    }
+}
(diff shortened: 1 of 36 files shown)


glye:
Early review ping @dpobel @andrerom @bdunogier @yannickroger - this is mainly about the architecture. See first commit: https://github.com/ezsystems/ez-support-tools/commit/116a6d7290a316183e672762944d7c41cf84cd5c

D1 and D2 are debatable. For D1 I'm unsure about the advantages. D2 seems a good thing - do we have an existing parallell I can base it on?
Any other architectural concerns?

bdunogier:
I'm lost. You speak about multiple changes, but there's only one commit... please help : -)

glye:
@bdunogier D3 and D4 are trivial, so I committed them together. I have amended the commit to make that clear, and I'll keep the others in separate, described commits.

bdunogier:
> For D1 I'm unsure about the advantages.

Value objects are important in domain driven architectures. Which we more or less follow.
It provides readability, eases, BC, and it is overall more consistent with the rest of the system.

Connected to D2, I'm thinking that once we have value objects, we could dig a bit southward, and consider re-using the view system, already used by ContentView, and planned for usage by ContentEdit (part of https://github.com/ezsystems/repository-forms/pull/62).

It would cover the template + rendering parts, and would let you focus on the system info one. It would also implicitly give us easy ways to (let) override view templates & controllers for customization.

glye:
>  I'm otoh not 100% sure if I'd use the same generic object, or a custom one per info. Both choices have consequences.

Do tell, @bdunogier :)

bdunogier:
> Do tell, @bdunogier :)

Custom properties/methods dictate the contract (interface), and make it easier to maintain, document and use.

Generic properties (hash) is more flexible, but only requires one value object.

I'd say it depends on semantic the sysinfos are meant/supposed to be (now, and in the future). Should we want to do anything special with any of them, then I'd go for more semantic => one ValueObject per Info item.

glye:
@bdunogier As far as has been discussed, no special actions have been mentioned. They need different templates for browser rendering, but that's would be handled by Views and is not relevant here, right? So that would indicate single value object.

Otoh a generic value object would just be a wrapper for the array that each info provider returns now, right? While a specific one for each could have a property for each of the current array keys. So if readability/documentation is a goal, that would indicate specific values...?

glye:
Added per-provider value objects (the controller isn't updated yet).

bdunogier:
Could you please not amend / rebase during review ? Having new commits for review changes makes it easier to check what has changed.

glye:
@bdunogier 
> Could you please not amend / rebase during review?
Noted. In this case I only changed the commit message to make it clear what it contained.

> Do you need one, given that it does nothing?
Not now, maybe later, and then it might be good for BC to have it in place. If you want it gone, say the word.

glye:
@bdunogier @dpobel Any comment on how best to inject the (variable number of) infoproviders into the controller? The current variadic isn't legit, is it? :)
https://github.com/ezsystems/ez-support-tools/blob/master/Controller/SystemInfoController.php#L22
https://github.com/ezsystems/ez-support-tools/blob/master/Resources/config/services.yml#L29

glye:
@bdunogier 
> Connected to D2, I'm thinking that once we have value objects, we could dig a bit southward, and consider re-using the view system, already used by ContentView

So... make a view that uses `eZ\Publish\Core\MVC\Symfony\View\BaseView` then, in a similar way to ContentView?

PR #4, opened 2016-03-11
Fix EZP-25564: Reimplement PHP system info collector

glye:
> Implements https://jira.ez.no/browse/EZP-25564, subtask of https://jira.ez.no/browse/EZP-25563
> Status: Ready for review

Collects information about PHP version and accelerator.

### DIFF
--- a/SystemInfo/Value/PhpSystemInfo.php
+++ b/SystemInfo/Value/PhpSystemInfo.php
@@ -18,14 +18,43 @@ class PhpSystemInfo extends ValueObject implements SystemInfo
     /**
      * PHP version.
      *
+     * Example: 5.6.7-1
+     *
      * @var string
      */
-    public $phpVersion;
+    public $version;
 
     /**
-     * PHP accelerator.
+     * True if an accelerator is enabled.
      *
-     * @var mixed
+     * @var bool
      */
-    public $phpAccelerator;
+    public $acceleratorEnabled;
+
+    /**
+     * PHP accelerator name, or null if no accelerator is enabled.
+     *
+     * Example: Zend OPcache
+     *
+     * @var string|null
+     */
+    public $acceleratorName;
+
+    /**
+     * PHP accelerator URL, or null if no accelerator is enabled.
+     *
+     * Example: http://www.php.net/opcache
+     *
+     * @var string|null
+     */
+    public $acceleratorURL;
+
+    /**
+     * PHP accelerator version, or null if no accelerator is enabled.
+     *
+     * Example: 7.0.4-devFE
+     *
+     * @var string|null
+     */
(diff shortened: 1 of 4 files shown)


SystemInfo/Value/PhpSystemInfo.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    /**
+     * PHP version.
+     *
+     * @var string
+     */
+    public $phpVersion;
+
+    /**
+     * PHP accelerator.
```
bdunogier: Misses description of array keys we expect in this property. An example would help.

SystemInfo/Value/PhpSystemInfo.php
```diff
@@ -0,0 +1,31 @@
...
+ * Value for information about the PHP interpreter and accelerator (if any) we are using.
+ */
+class PhpSystemInfo extends ValueObject implements SystemInfo
+{
+    /**
+     * PHP version.
+     *
+     * @var string
+     */
+    public $phpVersion;
```
bdunogier: I'd remove the "php" part, we are already in a `PhpSystemInfo`.

SystemInfo/Value/PhpSystemInfo.php
```diff
@@ -0,0 +1,31 @@
...
+
+use eZ\Publish\API\Repository\Values\ValueObject;
+
+/**
+ * Value for information about the PHP interpreter and accelerator (if any) we are using.
+ */
+class PhpSystemInfo extends ValueObject implements SystemInfo
+{
+    /**
+     * PHP version.
```
bdunogier: An example can't hurt, it specifies how the version is formatted.

SystemInfo/Value/PhpSystemInfo.php
```diff
@@ -0,0 +1,31 @@
...
+     * @var string
+     */
+    public $phpVersion;
+
+    /**
+     * PHP accelerator.
+     *
+     * @var mixed
+     */
+    public $phpAccelerator;
```
bdunogier: I'd prefer if we used separate properties for the different accelerator sub-properties. It would avoid the phpdoc/example issue, and it's not as if we had two dozens already.

glye:
Added property examples.
`phpVersion` -> `version`
Moved accelerator subproperties to top level.
Made clear which properties may be `null`.

SystemInfo/Value/PhpSystemInfo.php
glye: Added.

bdunogier:
+1.

glye:
Review ping @dpobel @yannickroger

dpobel:
+1

PR #7, opened 2016-03-14
Fix EZP-25566: Composer packages system info collector

glye:
> Implements https://jira.ez.no/browse/EZP-25566, subtask of https://jira.ez.no/browse/EZP-25563
> Status: Ready for review

Collects information about installed Composer packages.

- [x] Throw exception when composer file not found
- [x] ~~Complete move to using Composer classes~~ Nope, we won't do this
- [x] squash before merge if needed

### DIFF
--- a/Resources/config/services.yml
+++ b/Resources/config/services.yml
@@ -1,4 +1,5 @@
 parameters:
+    support_tools.info_collectors.composer.lock_file.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\JsonComposerLockSystemInfoCollector
     support_tools.info_collectors.database.doctrine.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\DoctrineDatabaseSystemInfoCollector
     support_tools.info_collectors.hardware.ezc.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\EzcHardwareSystemInfoCollector
     support_tools.info_collectors.php.ezc.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\EzcPhpSystemInfoCollector
@@ -7,6 +8,11 @@ services:
     support_tools.info_collector.system_info.ezc:
         class: ezcSystemInfo
         factory: [EzSystems\EzSupportToolsBundle\SystemInfo\EzcSystemInfoFactory, buildEzcSystemInfo]
+
+    support_tools.info_collectors.composer.lock_file:
+        class: %support_tools.info_collectors.composer.lock_file.class%
+        arguments:
+            - "%kernel.root_dir%/../composer.lock"
 
     support_tools.info_collectors.database.doctrine:
         class: %support_tools.info_collectors.database.doctrine.class%
--- a/SystemInfo/Collector/JsonComposerLockSystemInfoCollector.php
+++ b/SystemInfo/Collector/JsonComposerLockSystemInfoCollector.php
@@ -1,0 +1,82 @@
+<?php
+
+/**
+ * File containing the JsonComposerLockSystemInfoCollector class.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzSupportToolsBundle\SystemInfo\Collector;
+
+use EzSystems\EzSupportToolsBundle\SystemInfo\Exception;
+use EzSystems\EzSupportToolsBundle\SystemInfo\Value;
+
+/**
+ * Collects information about installed Composer packages, by reading json from composer.lock.
+ */
+class JsonComposerLockSystemInfoCollector implements SystemInfoCollector
+{
+    /**
+     * @var array Hash of stability constant values to human readable stabilities, see Composer\Package\BasePackage.
+     *
+     * Needed as long as we don't want to depend on Composer.
+     */
+    private $stabilities = [
+        0 => 'stable',
+        5 => 'RC',
+        10 => 'beta',
+        15 => 'alpha',
+        20 => 'dev',
+    ];
+
+    /**
+     * @var string Composer lock file with full path.
+     */
+    private $lockFile;
+
+    public function __construct($lockFile)
+    {
+        $this->lockFile = $lockFile;
+    }
+
+    /**
+     * Builds information about installed composer packages.
+     *
+     * @throws Exception\ComposerLockFileNotFoundException if the composer.lock file was not found.
+     *
+     * @return Value\ComposerSystemInfo
+     */
+    public function build()
+    {
+        if (!file_exists($this->lockFile)) {
+            throw new Exception\ComposerLockFileNotFoundException($this->lockFile);
+        }
+
+        $packages = [];
+        $lockData = json_decode(file_get_contents($this->lockFile), true);
+        foreach ($lockData['packages'] as $packageData) {
+            $packages[$packageData['name']] = new Value\ComposerPackage([
+                'name' => $packageData['name'],
+                'version' => $packageData['version'],
+                'dateTime' => new \DateTime($packageData['time']),
+                'homepage' => isset($packageData['homepage']) ? $packageData['homepage'] : '',
+                'reference' => $packageData['source']['reference'],
+            ]);
+
+            if (isset($lockData['stability-flags'][$packageData['name']])) {
+                $stabilityFlag = (int)$lockData['stability-flags'][$packageData['name']];
+
+                if (isset($this->stabilities[$stabilityFlag])) {
+                    $packages[$packageData['name']]->stability = $this->stabilities[$stabilityFlag];
+                }
+            }
+        }
+
+        ksort($packages, SORT_FLAG_CASE | SORT_STRING);
+
+        return new Value\ComposerSystemInfo([
+            'packages' => $packages,
+            'minimumStability' => isset($lockData['minimum-stability']) ? $lockData['minimum-stability'] : null,
+        ]);
+    }
+}
--- a/SystemInfo/Value/ComposerPackage.php
+++ b/SystemInfo/Value/ComposerPackage.php
@@ -1,0 +1,69 @@
+<?php
+
+/**
+ * File containing the ComposerPackage class.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzSupportToolsBundle\SystemInfo\Value;
+
+use eZ\Publish\API\Repository\Values\ValueObject;
+
+/**
+ * Value for information about a Composer package.
+ */
+class ComposerPackage extends ValueObject implements SystemInfo
+{
+    /**
+     * Name.
+     *
+     * Example: ezsystems/ezpublish-kernel
+     *
+     * @var string
+     */
+    public $name;
+
+    /**
+     * Version.
+     *
+     * Example: v2.7.10
+     *
+     * @var string
+     */
+    public $version;
+
+    /**
+     * Stability.
+     *
+     * One of: stable, RC, beta, alpha, dev, or null if not set.
+     *
+     * @var string|null
+     */
+    public $stability;
+
+    /**
+     * Date and time.
+     *
+     * @var \DateTime
+     */
+    public $dateTime;
+
+    /**
+     * Homepage URL, or null of not set.
+     *
+     * Example: https://symfony.com
+     *
+     * @var string|null
+     */
+    public $homepage;
+
+    /**
+     * Reference.
+     *
+     * Example: 9a3b6bf6ebee49370aaf15abc1bdeb4b1986a67d
+     *
+     * @var string
+     */
+    public $reference;
+}
--- a/SystemInfo/Value/ComposerSystemInfo.php
+++ b/SystemInfo/Value/ComposerSystemInfo.php
@@ -1,0 +1,35 @@
+<?php
+
+/**
+ * File containing the ComposerSystemInfo class.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzSupportToolsBundle\SystemInfo\Value;
+
+use eZ\Publish\API\Repository\Values\ValueObject;
+
+/**
+ * Value for information about installed Composer packages.
+ */
+class ComposerSystemInfo extends ValueObject implements SystemInfo
+{
+    /**
+     * Packages.
+     *
+     * A hash of composer package names and ComposerPackage values, or null if packages cannot be read.
+     *
+     * @var ComposerPackage[]|null
+     */
+    public $packages;
+
+    /**
+     * Minimum stability, as read from composer.lock.
+     *
+     * One of: stable, RC, beta, alpha, dev, or null if not set.
+     *
+     * @var string|null
+     */
+    public $minimumStability;
+}
--- a/Tests/SystemInfo/Collector/JsonComposerLockSystemInfoCollectorTest.php
+++ b/Tests/SystemInfo/Collector/JsonComposerLockSystemInfoCollectorTest.php
@@ -1,0 +1,78 @@
+<?php
+
+/**
+ * File containing the JsonComposerLockSystemInfoCollectorTest class.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzSupportToolsBundle\Tests\SystemInfo\Collector;
+
+use EzSystems\EzSupportToolsBundle\SystemInfo\Collector\JsonComposerLockSystemInfoCollector;
+use EzSystems\EzSupportToolsBundle\SystemInfo\Value\ComposerPackage;
+use EzSystems\EzSupportToolsBundle\SystemInfo\Value\ComposerSystemInfo;
+use PHPUnit_Framework_TestCase;
+
+class JsonComposerLockSystemInfoCollectorTest extends PHPUnit_Framework_TestCase
+{
+    /**
+     * @covers \EzSystems\EzSupportToolsBundle\SystemInfo\Collector\JsonComposerLockSystemInfoCollector::build()
+     */
+    public function testBuild()
+    {
+        $expected = new ComposerSystemInfo([
+            'packages' => [
+                'ezsystems/ezpublish-kernel' => new ComposerPackage([
+                    'name' => 'ezsystems/ezpublish-kernel',
+                    'version' => 'dev-master',
+                    'stability' => 'dev',
+                    'dateTime' => new \DateTime('2016-02-28 14:30:53'),
+                    'homepage' => 'http://share.ez.no',
+                    'reference' => 'ec897baa77c63b745749acf201e85b92bd614723',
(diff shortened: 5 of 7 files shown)


glye:
Fixed fixtures path which was breaking test.

andrerom:
I'm unsure on the exception case, there are cases where you might not have a lock file, however it should at the least generate a warning me thinks. /cc @bdunogier

glye:
Sounds good to me, warning trigger added.

Tests/SystemInfo/Collector/ComposerSystemInfoCollectorTest.php
```diff
@@ -0,0 +1,64 @@
...
+class ComposerSystemInfoCollectorTest extends PHPUnit_Framework_TestCase
+{
+    /**
+     * @var ComposerSystemInfoCollector
+     */
+    private $composerCollector;
+
+    public function setUp()
+    {
+        $this->composerCollector = new ComposerSystemInfoCollector(__DIR__ . '/_fixtures/');
```
bdunogier: If you wanna avoid flat files, I have used https://github.com/mikey179/vfsStream in the past. It will let you create a virtual FS (using streams). You would pass it as an argument to the collector, and add a `composer.lock` file from it. Not sure if better, but it is an option, and it doesn't rely on external files.

SystemInfo/Collector/ComposerSystemInfoCollector.php
```diff
@@ -0,0 +1,61 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzSupportToolsBundle\SystemInfo\Collector;
+
+use EzSystems\EzSupportToolsBundle\SystemInfo\Value;
+
+/**
+ * Collects information about installed Composer packages.
+ */
+class ComposerSystemInfoCollector implements SystemInfoCollector
```
bdunogier: What about `JsonComposerSystemInfoCollector` ? We might build a different implementation that uses a lib of some sorts.

SystemInfo/Value/ComposerSystemInfo.php
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Value for information about installed Composer packages.
+ */
+class ComposerSystemInfo extends ValueObject implements SystemInfo
+{
+    /**
+     * Packages.
+     *
+     * A two-dimensional hash of composer names and package information, or null if packages cannot be read.
```
bdunogier: Or it could use a `ComposerPackage` value object, which we can typehint on.

bdunogier:
Not sure in which case we wouldn't have a lock (besides tests), therefore not sure if a warning is actually better than an exception.

What we *could* do is make sure exceptions thrown by collectors are caught when appropriate. We probably don't want the whole support tools page to crash because _one_ collector had an issue.

SystemInfo/Collector/ComposerSystemInfoCollector.php
glye: Ok, or `JsonComposerLockSystemInfoCollector` to be concise, if neither json nor composer.lock are the only ways of getting this information.

SystemInfo/Value/ComposerSystemInfo.php
glye: Ok

Tests/SystemInfo/Collector/ComposerSystemInfoCollectorTest.php
glye: Nice to know, but also not sure if better. Should maybe have a unified scheme for all/most tests, and `_fixtures` seems most used.

glye:
Implemented package array as array of ComposerPackage value objects, and rebased on master to fix conflict.

glye:
Renamed `ComposerSystemInfoCollector` to `JsonComposerLockSystemInfoCollector`

SystemInfo/Collector/JsonComposerLockSystemInfoCollector.php
```diff
@@ -0,0 +1,61 @@
...
+ * Collects information about installed Composer packages, by reading json from composer.lock.
+ */
+class JsonComposerLockSystemInfoCollector implements SystemInfoCollector
+{
+    /**
+     * @var string Installation root directory
+     */
+    private $installDir;
+
+    public function __construct($installDir)
```
bdunogier: Why did you choose to use the directory as the argument, as opposed to the complete path to a lock file ?

SystemInfo/Value/ComposerPackage.php
```diff
@@ -0,0 +1,53 @@
...
+class ComposerPackage extends ValueObject implements SystemInfo
+{
+    /**
+     * Version.
+     *
+     * Example: v2.7.10
+     *
+     * @var string
+     */
+    public $version;
```
bdunogier: Do we want to distinguish stability, or expose it in a distinct property maybe ?

bdunogier:
No `name` property in the `ComposerPackage` value object ?

SystemInfo/Collector/JsonComposerLockSystemInfoCollector.php
glye: That's how it used to be. Makes sense to change it, now that you mention it.

SystemInfo/Value/ComposerPackage.php
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Date and time string.
+     *
+     * Example: 2016-02-28 20:37:19
+     *
+     * @var string
+     */
+    public $dateTimeString;
```
glye: Is it okay to keep this as a string? Or should we convert it, to be more flexible in how we display it?

bdunogier: I'd parse it to a `DateTime` without hesitation, good catch.

Resources/config/services.yml
```diff
@@ -14,6 +15,11 @@ services:
         arguments:
             - @support_tools.info_collector.system_info.ezc_system_info
 
+    support_tools.info_collectors.json_composer_lock:
```
bdunogier: Naming (same that for php and ezc): `support_tools.info_collectors.composer.lock_file`.

andrerom:
Note: If we are going to abstract composer packages like this then we jan just as well reuse this from composer itself. it has classes and functions for all this.

SystemInfo/Value/ComposerPackage.php
glye: How? Check if starting with "(v)0" or "dev-"? These versions seem somewhat inconsistent, the regexp for it may require some testing & maintenance. But I'm not against it.

glye:
> No name property in the ComposerPackage value object?

May want to add it. If so, are you ok with also keeping it as key in the hash, for convenient lookups?

glye:
Renamed `support_tools.info_collectors.json_composer_lock` -> `support_tools.info_collectors.composer.lock_file`

glye:
> Note: If we are going to abstract composer packages like this then we jan just as well reuse this from composer itself, it will provide additional features while at it, but it does mean we will need to pull in composer as a dependency.

Makes sense to me, and certainly before tweaking the value classes further. @bdunogier?

bdunogier:
It is worth a try.

Le mar. 15 mars 2016 13:21, Gunnstein Lye <notifications@github.com> a
écrit :

glye:
@bdunogier @andrerom First POC of move to Composer classes, test will fail. Just wanted to air a design question:
Right now I take the `Composer\Package` I receive from `Composer\Repository\InstalledFilesystemRepository`, pass it through `Composer\Package\Dumper\ArrayDumper` which gives me a hash of package properties, and fill my own `ComposerPackage` value object with that. Which works fine, apart from having to implement the ca. 40 existing properties, and the fact that I'm actually doing what we wanted to avoid: Reimplementing the composer api.

OTOH, `Composer\Package` is not a value object and we kind of need it to be. So... should the `ComposerPackage` value object implement `Composer\Package\PackageInterface`, and we implement a repo that inherits from `InstalledFilesystemRepository`, ensuring it returns our `ComposerPackage` instead of Composer's `Package`?

Or, should our `ComposerPackage` be a wrapper for `Composer\Package`? Would it be possible to reimplement the getter/setter to work on the underlying properties? This sounds better to me. Will get back to it in the morning!

Ref: https://getcomposer.org/apidoc/master/index.html

andrerom:
Ok, might look like a fork we might don't want to walk down then, seems a bit overkill to implement all this. /c @bdunogier ?

glye:
I'm working on the last alternative now and it doesn't look bad to me. It seems fairly easy to overload the relevant ValueObject methods to work on the getter methods of an internal Composer PackageInterface object. If the idea isn't utterly haram to you guys, I'd like to give it a few more hours for a working POC.

bdunogier:
I haven't fully read your comment, @glye, but to "seems a bit overkill to implement all this": yes, I think I already agree. I'll comment further when I'm done reading.

bdunogier:
>  It seems fairly easy to overload the relevant ValueObject methods to work on the getter methods

I'm not sure we want to do that. We have be careful to keep stateless objects.

glye:
Last commit shows the last approach suggested (test failing due to a "class not found" I don't understand yet).
- `ComposerSystemInfo` value has an array of `ComposerPackage` values, as before.
- `ComposerPackage` value has a single property, an internal `Composer\Package\PackageInterface`. The `__get()` and `getProperties()` methods are reimplemented to work on the properties of the internal object. A `publicProperties()` method is added because `getProperties()` is protected and `attributes()` is final - it's handy for the dumper command to get a list of all available properties. (See also `toArray()` below.)
- `ComposerInstalledFileSystemInfoCollector` is now simpler, it just wraps the composer package object inside our value object.
- `SystemInfoDumpCommand` had to be extended so it works when a valueobject contains an array of value objects. (We could add a `toArray()` method in our value objects instead.)
- The composer.lock fixture is replaced with installed.json, because that's what the Composer classes expects.

@bdunogier @andrerom This seems clean to me, apart from TODOs in the code.

bdunogier:
I'd be curious to hear from @andrerom about the stateless objects.

bdunogier:
Overall, I'm also wondering if we should add composer as a _requirement_.

SystemInfo/Value/ComposerPackage.php
```diff
@@ -0,0 +1,86 @@
...
+     *
+     * @param array $dynamicProperties Additional dynamic properties exposed on the object
+     *
+     * @return array
+     */
+    protected function getProperties($dynamicProperties = array())
+    {
+        $properties = $dynamicProperties;
+
+        // TODO Not very nice. Use hardcoded lookup table instead?
```
andrerom: yes, we don't want to expose everything I think

andrerom:
> Also, I'm also wondering if we should add composer as a requirement.

like it is now?

> The composer.lock fixture is replaced with installed.json, because that's what the Composer classes expect.

from the looks of the fixure this is just data from composer.json or? so where does it get the installed version info from? Is it also reading lock file, or does it do a full composer update lookup to give you versions of latests packages *(something to avoid at all cost, to slow :) )*.

glye:
> from the looks of the fixure this is just data from composer.json or? so where does it get the installed version info from? Is it also reading lock file, or does it do a full composer update lookup to give you versions of latests packages (something to avoid at all cost, to slow :) ).

The fixture is a subset of `vendor/composer/installed.json`. Afaik, composer updates this file on every composer install/update. The collector only reads this static file, it's fast.

glye:
@andrerom I proposed to @bdunogier to copy the branch as it is now, then revert to when we didn't use the Composer API. Due to problems mentioned it may be complex to finish the current approach, but it's something to keep in mind for later. OK with you?

bdunogier:
(need to update master after the naming PR merge)

andrerom:
> OK with you?

yes, go ahead, seems to complex so lets revert and rather copy paste concepts we want from there :)

bdunogier:
What is left to do besides rebasing on top of latest master, @glye ?

glye:
Reverted approach using composer api, copy at https://github.com/ezsystems/ez-support-tools/tree/ezp-25566_composer_system_info_collector_using_composer_api
Rebased on master
Converted datetime string to DateTime
Use full path to lockfile, not just the directory

TODO use exception rather than warning when lockfile not found
(recheck feedback for other issues)

glye:
Added name property in ComposerPackage value object.

SystemInfo/Collector/JsonComposerLockSystemInfoCollector.php
glye: Using complete file path now.

SystemInfo/Value/ComposerPackage.php
glye: Using DateTime now.

Resources/config/services.yml
glye: Done.

glye:
Throw exception when composer file not found - catching it will be done in the `SystemInfoViewBuilder` and in the real Command (not the current temporary dumper).

glye:
All issues taken care of afaik, ready for review @bdunogier @andrerom

SystemInfo/Value/ComposerPackage.php
andrerom: what kind of data do you have to pick from? at least in composer.lock there is info on stability and normalizes versions.

andrerom: But @bdunogier has a point, there should be a separate property for stability flag, this is about not how stable the package is, but which stability channel we pull in new packages on.

_But_ I think you'll only find this in composer.lock, like this:
```
    "stability-flags": {
        "ezsystems/ezpublish-kernel": 20,
        "ezsystems/repository-forms": 20,
        "ezsystems/ezplatform-solr-search-engine": 20,
        "ezsystems/platform-ui-bundle": 20,
        "ezsystems/behatbundle": 20
    },
```

20 here refers to `@dev` as set in composer.json in master branch of ezplatform.

SystemInfo/Value/ComposerPackage.php
```diff
@@ -0,0 +1,60 @@
...
+     * @var \DateTime
+     */
+    public $dateTime;
+
+    /**
+     * Homepage URL.
+     *
+     * Example: https://symfony.com
+     *
+     * @var string
```
andrerom: `string|null` it is not always set

SystemInfo/Value/ComposerPackage.php
glye: Ok! We're back to using composer.lock now, so the data is available.

Tests/SystemInfo/Collector/JsonComposerLockSystemInfoCollectorTest.php
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * @var JsonComposerLockSystemInfoCollector
+     */
+    private $composerCollectorNotFound;
+
+    public function setUp()
+    {
+        $this->composerCollector = new JsonComposerLockSystemInfoCollector(__DIR__ . '/_fixtures/composer.lock');
+        $this->composerCollectorNotFound = new JsonComposerLockSystemInfoCollector(__DIR__ . '/_fixtures/snafu.lock');
```
bdunogier: I wouldn't have bothered with creating private vars, given that they're only used once each.

bdunogier:
+1

glye:
Added stability-flags property reading from composer.lock - didn't immediately see how to convert e.g. "20" to "dev", found no mention of it in composer.json. Will see if it can be improved.
Made homepage url optional
Removed private vars in test

SystemInfo/Value/ComposerPackage.php
```diff
@@ -34,6 +34,16 @@ class ComposerPackage extends ValueObject implements SystemInfo
     public $version;
 
     /**
+     * Stability.
+     *
+     * String with integer, as set in composer.lock "stability-flags".
```
bdunogier: I would try to provide a string instead of an integer value.

SystemInfo/Value/ComposerPackage.php
glye: ...though I didn't see a way to get from "20" to "dev" yet.

SystemInfo/Value/ComposerPackage.php
glye: Of course, working on it as mentioned.

glye:
The stability property of ComposerPackage is now a human readable string: stable, RC, beta, alpha, dev, or `null`. These are defined as constants in `Composer\Package\BasePackage`, but if we're not going to depend on Composer we can use that, hence this hack. (I'm open for improvements suggestions, or just going back to previous commit.)

Also, I could read `minimum-stability` from composer.lock and use that instead of `null` when no specific stability is set, as that is the default stability.
@andrerom @bdunogier

SystemInfo/Value/ComposerPackage.php
bdunogier: Sorry, wrong review order :-)

P.S. if you want to avoid complicated testing, I suggest you use a dedicated service for parsing the stability.

glye: Well, the tests are not complicated by this at the moment, though I am sure we can find ways to make them so :)
But first let's see if the stability feature/hack is something you guys want, given the hardcoding.

andrerom:
> Also, I could read minimum-stability from composer.lock and use that instead of null when no specific stability is set, as that is the default stability.

True, we need to keep track of that somehow, I see two options:
A. If we plan to dump full composer.json and .lock into debug tools anyway, we can indeed simplify like you imply, taking value from minimal-stability when not set.
B. If not, then we need to know in debug reports if stability comes form package or global minimal-stability in order to give proper support suggestions. If so we can add `minimumStability` on `Value\ComposerSystemInfo` based on `minimum-stability` in lock.


wdt @glye @bdunogier ?  *(I might prefer option B as even in UI it might be good to have separation of these two concepts)*

bdunogier:
>  If so we can add minimumStability on Value\ComposerSystemInfo based on minimum-stability in lock.

Nice :-) I'm okay with solution B. It's not gonna kill anybody, in any case (if it does, I'm interested in the use-cases).

glye:
Added @andrerom's option B with minimumStability read from composer.lock.

SystemInfo/Collector/JsonComposerLockSystemInfoCollector.php
```diff
@@ -54,8 +54,10 @@ public function build()
 
         $packages = [];
         $lockData = json_decode(file_get_contents($this->lockFile), true);
+        $minimumStability = isset($lockData['minimum-stability']) ? $lockData['minimum-stability'] : null;
+
         foreach ($lockData['packages'] as $packageData) {
-            $stability = null;
+            $stability = $minimumStability;
```
andrerom: this is not B :)

glye: It's a synthesis of both, it seems :) Ok, lets B-ify it.

glye:
Option B take 2, this should be it. #FamousLastWords

andrerom:
+1

SystemInfo/Collector/JsonComposerLockSystemInfoCollector.php
bdunogier: I actually disagree... I don't see why the UI would need to do anything about it. 

The composer doc says that `minimum-stability` is stable unless specified otherwise. It seems to me that it is the responsibility of this collector to handle this.

glye: @bdunogier Follow-up to this at https://jira.ez.no/browse/EZP-25608

PR #8, opened 2016-03-14
EZP-25579: SystemInfo view

bdunogier:
> https://jira.ez.no/browse/EZP-25579
> Working prototype, with all the hard requirements
> PlatformUI integration: https://github.com/ezsystems/PlatformUIBundle/pull/532
> ezplatform integration: https://github.com/ezsystems/ezplatform/compare/ezp25579-view_api_integration

### Contents
- `SystemInfoViewController` + `viewInfoAction`
- `/ez-support-tools/view/info/{identifier}` route. Identifier is the name of the class minus SystemInfo, underscorized: `PhpSystemInfo` :arrow_right: `php`
- `SystemInfoView` object + builder
- `SystemInfo\Identifier` view matcher

### Status
`/ez-support-tools/view/php` matches the SystemInfoController view action.
The template (property of the view) used to render the SystemInfoView is set depending on the view matchers configuration.

#### View configuration
The template used by each info type can be configured as for  `content_view`:

```
parameters:
    ezsettings.default.system_info_view:
        full:
            php:
                template: "ez-support-tools/info/php.html.twig"
                match:
                    SystemInfo\Identifier: "php"
            hardware:
                template: "ez-support-tools/info/hardware.html.twig"
                match:
                    SystemInfo\Identifier: "hardware"
```

#### View templates
The viewed `SystemInfo` object is provided as the `info` variable.

We don't intend to provide system info outside of a PlatformUI context for the time being. Therefore, the *pjax* view templates should be added to PlatformUI.

#### Permissions & security

How was this handled in PlatformUI ?
First, I think we can _maybe_ postpone that part: the package is still a 0.x, and we can easily document that in the readme.

As for implementing this, I'd lean towards real policies: we can have services around collectors that do check permissions using custom policies.

### Tasks
- [x] PlatformUI templates. The idea is that PlatformUI would use its own templates to implement pjax
- [x] Tests
- [x] Semantic config for system_info_view. But is it a hard requirement ?
- [x] Think about security/permissions

### DIFF
--- a/Controller/SystemInfoController.php
+++ b/Controller/SystemInfoController.php
@@ -1,0 +1,18 @@
+<?php
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzSupportToolsBundle\Controller;
+
+use EzSystems\EzSupportToolsBundle\View\SystemInfoView;
+
+/**
+ * Do we need that controller at all ?
+ */
+class SystemInfoController
+{
+    public function viewInfoAction(SystemInfoView $view)
+    {
+        return $view;
+    }
+}
--- a/Resources/config/services.yml
+++ b/Resources/config/services.yml
@@ -25,7 +25,10 @@ services:
 
     support_tools.view.system_info_view_builder:
         class: EzSystems\EzSupportToolsBundle\View\SystemInfoViewBuilder
-        arguments: [@support_tools.info_collectors.ezc_hardware]
+        arguments:
+            -
+                hardware: @support_tools.info_collectors.ezc_hardware
+                php: @support_tools.info_collectors.php
 
     support_tools.view.controller:
         class: EzSystems\EzSupportToolsBundle\Controller\SystemInfoController
--- a/Resources/config/services.yml
+++ b/Resources/config/services.yml
@@ -22,3 +22,10 @@ services:
         class: %support_tools.info_collectors.php.ezc.class%
         arguments:
             - @support_tools.info_collector.system_info.ezc
+
+    support_tools.view.system_info_view_builder:
+        class: EzSystems\EzSupportToolsBundle\View\SystemInfoViewBuilder
+        arguments: [@support_tools.info_collectors.ezc_hardware]
+
+    support_tools.view.controller:
+        class: EzSystems\EzSupportToolsBundle\Controller\SystemInfoController
--- a/Resources/config/services.yml
+++ b/Resources/config/services.yml
@@ -1,3 +1,6 @@
+imports:
+    - { resource: view.yml }
+
 parameters:
     support_tools.info_collectors.database.doctrine.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\DoctrineDatabaseSystemInfoCollector
     support_tools.info_collectors.hardware.ezc.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\EzcHardwareSystemInfoCollector
@@ -22,13 +25,3 @@ services:
         class: %support_tools.info_collectors.php.ezc.class%
         arguments:
             - @support_tools.info_collector.system_info.ezc
-
-    support_tools.view.system_info_view_builder:
-        class: EzSystems\EzSupportToolsBundle\View\SystemInfoViewBuilder
-        arguments:
-            -
-                hardware: @support_tools.info_collectors.ezc_hardware
-                php: @support_tools.info_collectors.php
-
-    support_tools.view.controller:
-        class: EzSystems\EzSupportToolsBundle\Controller\SystemInfoController
--- a/View/SystemInfoView.php
+++ b/View/SystemInfoView.php
@@ -1,0 +1,41 @@
+<?php
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzSupportToolsBundle\View;
+
+use eZ\Publish\Core\MVC\Symfony\View\BaseView;
+use eZ\Publish\Core\MVC\Symfony\View\View;
+
+class SystemInfoView extends BaseView implements View
+{
+    /**
+     * @var \EzSystems\EzSupportToolsBundle\SystemInfo\Value\SystemInfo
+     */
+    private $info;
+
+    /**
+     * @param \EzSystems\EzSupportToolsBundle\SystemInfo\Value\SystemInfo $info
+     *
+     * @return SystemInfoView
+     */
+    public function setInfo($info)
+    {
+        $this->info = $info;
+
+        return $this;
+    }
+
+    /**
+     * @return \EzSystems\EzSupportToolsBundle\SystemInfo\Value\SystemInfo
+     */
+    public function getInfo()
+    {
+        return $this->info;
+    }
+
+    protected function getInternalParameters()
+    {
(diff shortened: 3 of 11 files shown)


Resources/config/services.yml
```diff
@@ -16,3 +16,10 @@ services:
     support_tools.info_collector.system_info.ezc_system_info:
         class: ezcSystemInfo
         factory: [EzSystems\EzSupportToolsBundle\SystemInfo\EzcSystemInfoFactory, buildEzcSystemInfo]
+
+    support_tools.view.system_info_view_builder:
+        class: EzSystems\EzSupportToolsBundle\View\SystemInfoViewBuilder
+        arguments: [@support_tools.info_collectors.ezc_hardware]
```
bdunogier: Note to self: probably not final.

View/SystemInfoView.php
```diff
@@ -0,0 +1,41 @@
+<?php
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzSupportToolsBundle\View;
+
+use eZ\Publish\Core\MVC\Symfony\View\BaseView;
+use eZ\Publish\Core\MVC\Symfony\View\View;
+
+class SystemInfoView extends BaseView implements View
```
bdunogier: Note to self: phpdoc.

View/SystemInfoViewBuilder.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        return $argument === 'support_tools.view.controller:viewInfoAction';
+    }
+
+    public function buildView(array $parameters)
+    {
+        $view = new SystemInfoView();
+        $view->setInfo($this->infoCollector->build());
+        $view->setTemplateIdentifier(
+            $this->toTemplateIdentifier($view->getInfo())
```
bdunogier: Note to self: must be changed.

The template identifier should come from the ViewMatchers (to @glye : executed by the Configurator, in the View Request Listener).

It could be interesting to base the default template name on a convention, probably from the view builder (computed based on the class name, as it is now), with a convention based directory (`HardwareSystemInfo` => `ez-support-tools/hardware_system_info.html.twig`).

We are kind of getting close from the concept we had in legacy, where the _requested template_ was part of the view rule. We don't have this feature yet (matchers can't/don't test that).

Shorter version: if a default template is set by convention, based on the rendered info, in the view object, how could we take advantage of that in view rules ?

View/SystemInfoViewBuilder.php
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzSupportToolsBundle\View;
+
+use Doctrine\Common\Inflector\Inflector;
+use eZ\Publish\Core\MVC\Symfony\View\Builder\ViewBuilder;
+use EzSystems\EzSupportToolsBundle\SystemInfo\Collector\SystemInfoCollector;
+
+class SystemInfoViewBuilder implements ViewBuilder
```
bdunogier: Should probably also implement `CachableView`.

glye: Agree, but.. how should this cache be expired?

View/SystemInfoViewBuilder.php
glye: > if a default template is set by convention, based on the rendered info, in the view object, how could we take advantage of that in view rules?

Isn't that answered by `toTemplateIdentifier()`? It's somewhat foggy to me, though I'm all for convention rather than configuration.

View/SystemInfoViewBuilder.php
bdunogier: Do we need to expire it, as long as there is a TTL ?

View/SystemInfoViewBuilder.php
bdunogier: Well, so far we _only_ did configuration.

The `toTemplateIdentifier()` indeed does that. On the other hand, it's hack, and it's not done like that at all in the places where we use this API.

Controller/SystemInfoController.php
```diff
@@ -0,0 +1,18 @@
+<?php
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzSupportToolsBundle\Controller;
+
+use EzSystems\EzSupportToolsBundle\View\SystemInfoView;
+
+/**
+ * Do we need that controller at all ?
```
bdunogier: Needs answer before merge.

bdunogier:
The critical path should be working now. Updating the description with current state and how to use.

bdunogier:
Added some stuff about permissions & security, with questions.

Controller/SystemInfoController.php
```diff
@@ -0,0 +1,18 @@
...
+ */
+class SystemInfoController
+{
+    public function viewInfoAction(SystemInfoView $view)
```
glye: Isn't this parameter a string - `systemInfoIdentifier` as defined in routing.yml? Or is some automagic taking care of it?

glye:
> Create the matching template in app/Resources/view. Example: app/Resources/ez-support-tools/info/hardware.html.twig.

`app/Resources/view/ez-support-tools/info/hardware.html.twig` then..?
(Alternatively they can be put in the bundle, right?)

bdunogier:
> (Alternatively they can be put in the bundle, right?)

Yes, indeed. But I don't think we should right now, because of scope.
Our scope is to provide this feature withinPlatformUI / pjax. We don't plan on providing SystemInfo outside of this context.

I'll push very soon override templates _in PlatformUI_, it will make more sense :-)

Controller/SystemInfoController.php
bdunogier: Automagic it is: the "View API".

glye:
> How was this handled in PlatformUI?

It just requires the `setup/system_info` policy, it seems.

bdunogier:
> Did someone edit my comment? I don't recall the above line :)
Actually I did. Bad reflex, might have been (too) late. Hit 'edit' instead of 'reply'. The buttons are sooo close from each other :D

bdunogier:
About permissions, we should aim, as soon as possible, to having a service layer that includes permissions (could be a decorator).

This principle only exists in the Public API, but there are no real reasons why we'd do things differently in an extension. Now the question is what does this service _serve_ exactly ? The SystemInfo view isn't our end goal. It's only meant to be used to render the SystemInfo objects from a "Report". Should we build that service around reports, and use that in our View ?

glye:
Re: permissions, what granularity do we need? Per infocollector, per report, or a single one for the whole thing (probably not).

bdunogier:
> Re: permissions, what granularity do we need?

We can be agile about it. For now, we should just stick to what we used before.
Any opinion on the service layer aspect ?

glye:
I thought we'd need some idea about intended granularity before designing something that deals with permissions. Anyway, what else should the service layer do?

bdunogier:
> Anyway, what else should the service layer do?

It should do what we need it to do ? :-)

But before we dig into that, I suggest that we merge this PR without handling it, and create a follow-up before we merge.

bdunogier:
Created follow-ups for [permissions](https://jira.ez.no/browse/EZP-25614) and [implementation of PJAX templates in PlatformUI](https://jira.ez.no/browse/EZP-25615).

PR #11, opened 2016-03-23
Fix EZP-25599: Symfony kernel system info collector

glye:
> Implements https://jira.ez.no/browse/EZP-25599
> Status: Ready for review

Collects all Symfony kernel parameters.

### DIFF
--- a/Resources/config/services.yml
+++ b/Resources/config/services.yml
@@ -6,6 +6,7 @@ parameters:
     support_tools.info_collectors.database.doctrine.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\DoctrineDatabaseSystemInfoCollector
     support_tools.info_collectors.hardware.ezc.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\EzcHardwareSystemInfoCollector
     support_tools.info_collectors.php.ezc.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\EzcPhpSystemInfoCollector
+    support_tools.info_collectors.symfony.kernel.config.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\ConfigurationSymfonyKernelSystemInfoCollector
 
 services:
     support_tools.info_collector.system_info.ezc:
@@ -31,3 +32,9 @@ services:
         class: %support_tools.info_collectors.php.ezc.class%
         arguments:
             - @support_tools.info_collector.system_info.ezc
+
+    support_tools.info_collectors.symfony.kernel.config:
+        class: %support_tools.info_collectors.symfony.kernel.config.class%
+        arguments:
+            - @kernel
+            - %kernel.bundles%
--- a/SystemInfo/Collector/ConfigurationSymfonyKernelSystemInfoCollector.php
+++ b/SystemInfo/Collector/ConfigurationSymfonyKernelSystemInfoCollector.php
@@ -1,0 +1,68 @@
+<?php
+
+/**
+ * File containing the ConfigurationSymfonyKernelSystemInfoCollector class.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzSupportToolsBundle\SystemInfo\Collector;
+
+use EzSystems\EzSupportToolsBundle\SystemInfo\Value;
+use Symfony\Component\HttpKernel\Kernel;
+
+/**
+ * Collects information about the Symfony kernel we are using.
+ */
+class ConfigurationSymfonyKernelSystemInfoCollector implements SystemInfoCollector
+{
+    /**
+     * Symfony kernel.
+     *
+     * @var \Symfony\Component\HttpKernel\Kernel
+     */
+    private $kernel;
+
+    /**
+     * Installed bundles.
+     *
+     * A hash containing the active bundles, where the key is the bundle name, and the value is the corresponding namespace.
+     *
+     * Example:
+     * array (
+     *   'AppBundle' => 'AppBundle\\AppBundle',
+     *   'AsseticBundle' => 'Symfony\\Bundle\\AsseticBundle\\AsseticBundle',
+     * )
+     *
+     * @var array
+     */
+    private $bundles;
+
+    public function __construct(Kernel $kernel, array $bundles)
+    {
+        $this->kernel = $kernel;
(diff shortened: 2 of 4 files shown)


SystemInfo/Collector/SymfonySystemInfoCollector.php
```diff
@@ -0,0 +1,65 @@
...
+ */
+namespace EzSystems\EzSupportToolsBundle\SystemInfo\Collector;
+
+use EzSystems\EzSupportToolsBundle\SystemInfo\Value;
+use Symfony\Component\HttpKernel\Kernel;
+
+/**
+ * Collects information about the Symfony installation we are using.
+ */
+class SymfonySystemInfoCollector implements SystemInfoCollector
```
bdunogier: Naming naming...

It uses configuration data to collect symfony info. What about `ConfigurationSymfonySystemInfoCollector` ?

bdunogier:
Looks okay'ish. But based on what you ask about in the description, @glye, aren't we actually facing a `SymfonyKernelSystemInfo` ? It's kind of the same for now, but there is more to Symfony than just the kernel.

SystemInfo/Collector/SymfonySystemInfoCollector.php
glye: Since we only collect kernel data, it would be `ConfigurationSymfonyKernelSystemInfoCollector` then..?

glye:
@bdunogier Two reasons for the test to fail. I fixed 1) in https://github.com/ezsystems/ez-support-tools/commit/88409dbee7f17806064d00219e99aedb11c08aa0
but 2) is weird. I assume this passed before you pushed it, any idea why it would fail now? https://travis-ci.org/ezsystems/ez-support-tools/jobs/119215018

SystemInfo/Collector/ConfigurationSymfonyKernelSystemInfoCollector.php
```diff
@@ -0,0 +1,129 @@
...
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzSupportToolsBundle\SystemInfo\Collector;
+
+use EzSystems\EzSupportToolsBundle\SystemInfo\Value;
+use Symfony\Component\HttpKernel\Kernel;
+
+/**
+ * Collects information about the Symfony installation we are using.
```
bdunogier: "About the Symfony kernel" ?

SystemInfo/Collector/ConfigurationSymfonyKernelSystemInfoCollector.php
```diff
@@ -0,0 +1,129 @@
...
+     * @var string
+     */
+    private $charset;
+
+    /**
+     * Container class.
+     *
+     * @var string
+     */
+    private $containterClass;
```
bdunogier: Typo (extra `t`)

Resources/config/services.yml
```diff
@@ -31,3 +32,16 @@ services:
         class: %support_tools.info_collectors.php.ezc.class%
         arguments:
             - @support_tools.info_collector.system_info.ezc
+
+    support_tools.info_collectors.symfony.kernel.config:
+        class: %support_tools.info_collectors.symfony.kernel.config.class%
+        arguments:
+            - %kernel.environment%
```
bdunogier: That's a lot of arguments. What about injecting `%kernel%` instead ? It would also simplify the collector.

glye:
@bdunogier Done renaming, and it collects all kernel parameters now.

SystemInfo/Collector/ConfigurationSymfonyKernelSystemInfoCollector.php
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * True if Symfony is in debug mode.
+     *
+     * @var bool
+     */
+    private $debugMode;
+
+    /**
+     * An array containing the active bundles (keys) and the corresponding namespace.
```
bdunogier: You should specify that the key is the bundle's name. An example of an array row would also help, I think.

Resources/config/services.yml
glye: @bdunogier `%kernel%` isn't something that can be injected (non-existing parameter) but perhaps it can be used some other way. Digging...

glye: Ok, I can inject `@kernel` instead. Need to do more mocking in the test before I push it. I'd like to keep the `%kernel.bundles%` parameter, to avoid complicating the collector (I'd rather use the hash it provides directly, than build the same thing in collector code).

SystemInfo/Collector/ConfigurationSymfonyKernelSystemInfoCollector.php
glye: fixed

SystemInfo/Collector/ConfigurationSymfonyKernelSystemInfoCollector.php
glye: Right, though I am no longer collecting this data. It was the name of the class. Now, when I'm passing the whole kernel object to the collector, I get the container object instead, and I'd have to get the class name from that. Not hard, but it seems a rather obscure/pointless thing to deal with. Let me know if you want it back.

SystemInfo/Collector/ConfigurationSymfonyKernelSystemInfoCollector.php
glye: fixed

glye:
Feedback fixes: Doc, skipped containerClass, injecting only `@kernel` and `%kernel.bundles%`.

andrerom:
+1

SystemInfo/Collector/ConfigurationSymfonyKernelSystemInfoCollector.php
```diff
@@ -0,0 +1,68 @@
...
+     * array (
+     *   'AppBundle' => 'AppBundle\\AppBundle',
+     *   'AsseticBundle' => 'Symfony\\Bundle\\AsseticBundle\\AsseticBundle',
+     * )
+     *
+     * @var array
+     */
+    private $bundles;
+
+    public function __construct($kernel, $bundles)
```
bdunogier: Typehint on `$kernel` ?

bdunogier:
Great,  +1. You can go ahead and merge, @glye.

glye:
Thanks, added typehint and squashed.

PR #3, opened 2016-03-09
Fix EZP-25490: Reimplement SystemInfo in ez-support-tools

glye:
> Fixes https://jira.ez.no/browse/EZP-25490
> Status: Ready for review

Implements the EzcHardware info collector only (split out from EzcSystem info for compartmentalisation reasons) to verify architecture and naming, and a simple, temporary Command:

```console
php app/console ez-support-tools:dump-info support_tools.info_collectors.ezc_hardware
array (
  'cpuType' => 'Intel(R) Core(TM) i7-3720QM CPU @ 2.60GHz',
  'cpuSpeed' => 2591.9000000000001,
  'cpuCount' => 1,
  'memorySize' => 1554632704,
)
```
- [x] Testing

PR #12, opened 2016-03-31
Fix EZP-25598: eZSupportTools infocollector registry

glye:
> Implements https://jira.ez.no/browse/EZP-25598
> Status: Ready for review

This does not implement infocollector descriptions, that's not needed/important (at least for now).

- [x] Implement the registry
- [x] Update the dump command to work with the registry
- [x] Update the View to work with the registry
- [x] Make the dump command a service and inject the registry
- [x] Testing

### DIFF
--- a/Command/SystemInfoDumpCommand.php
+++ b/Command/SystemInfoDumpCommand.php
@@ -8,13 +8,30 @@
  */
 namespace EzSystems\EzSupportToolsBundle\Command;
 
+use EzSystems\EzSupportToolsBundle\SystemInfo\Collector\SystemInfoCollector;
+use EzSystems\EzSupportToolsBundle\SystemInfo\SystemInfoCollectorRegistry;
 use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
 use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
 class SystemInfoDumpCommand extends ContainerAwareCommand
 {
+    /**
+     * System info collector registry.
+     *
+     * @var \EzSystems\EzSupportToolsBundle\SystemInfo\SystemInfoCollectorRegistry
+     */
+    private $registry;
+
+    public function __construct(SystemInfoCollectorRegistry $registry)
+    {
+        $this->registry = $registry;
+
+        parent::__construct();
+    }
+
     /**
      * Define command and input options.
      */
@@ -23,10 +40,22 @@ class SystemInfoDumpCommand extends ContainerAwareCommand
         $this
             ->setName('ez-support-tools:dump-info')
             ->setDescription('Collects system information and dumps it.')
+            ->setHelp(<<<'EOD'
+By default it dumps information from all available information collectors.
+You can specify one or more collectors as arguments, e.g. 'php database hardware'.
+To get a list if available collectors, use '--list-info-collectors'
+EOD
+            )
+            ->addOption(
+                'list-info-collectors',
+                null,
+                InputOption::VALUE_NONE,
+                'List all available information collectors, and exit.'
+            )
             ->addArgument(
-                'info-collector',
-                InputArgument::REQUIRED,
-                'Which information collector should be used?'
+                'info-collectors',
+                InputArgument::IS_ARRAY,
+                'Which information collector(s) should be used? (separate multiple names with spaces)'
             )
             ;
     }
@@ -39,11 +68,37 @@ class SystemInfoDumpCommand extends ContainerAwareCommand
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $infoCollector = $this->getContainer()->get($input->getArgument('info-collector'));
+        if ($input->getOption('list-info-collectors')) {
+            $output->writeln('Available info collectors:', true);
+            foreach ($this->registry->getIdentifiers() as $identifier) {
+                $output->writeln("  $identifier", true);
+            }
+        } else if ($identifiers = $input->getArgument('info-collectors')) {
+            foreach ($identifiers as $identifier) {
+                $this->outputInfo(
+                    $this->registry->getItem($identifier),
+                    $output
+                );
+            }
+        } else {
+            foreach ($this->registry->getIdentifiers() as $identifier) {
+                $this->outputInfo($this->registry->getItem($identifier), $output);
+            }
+        }
+    }
+
+    /**
+     * Output info collected by the given collector.
+     *
+     * @param $infoCollector SystemInfoCollector
+     * @param $output OutputInterface
+     */
+    private function outputInfo(SystemInfoCollector $infoCollector, OutputInterface $output)
+    {
         $infoValue = $infoCollector->collect();
 
         $outputArray = [];
-        // attributes() is deprecated, and getProperties() is protected. Smeg it, this is very temporary anyway.
+        // attributes() is deprecated, and getProperties() is protected. TODO add a toArray() or similar.
         foreach ($infoValue->attributes() as $property) {
             $outputArray[$property] = $infoValue->$property;
         }
--- a/DependencyInjection/Compiler/SystemInfoCollectorPass.php
+++ b/DependencyInjection/Compiler/SystemInfoCollectorPass.php
@@ -21,20 +21,20 @@ class SystemInfoCollectorPass implements CompilerPassInterface
      */
     public function process(ContainerBuilder $container)
     {
-        if (!$container->has('support_tools.info_collector.registry')) {
+        if (!$container->has('support_tools.system_info.collector_registry')) {
             return;
         }
 
-        $infoCollectorsTagged = $container->findTaggedServiceIds('support_tools.info_collector');
+        $infoCollectorsTagged = $container->findTaggedServiceIds('support_tools.system_info.collector');
 
         $infoCollectors = [];
         foreach ($infoCollectorsTagged as $id => $tags) {
             foreach ($tags as $attributes) {
-                $infoCollectors[$attributes['alias']] = new Reference($id);
+                $infoCollectors[$attributes['identifier']] = new Reference($id);
             }
         }
 
-        $infoCollectorRegistryDef = $container->findDefinition('support_tools.info_collector.registry');
+        $infoCollectorRegistryDef = $container->findDefinition('support_tools.system_info.collector_registry');
         $infoCollectorRegistryDef->setArguments([$infoCollectors]);
     }
 }
--- a/DependencyInjection/Compiler/SystemInfoCollectorPass.php
+++ b/DependencyInjection/Compiler/SystemInfoCollectorPass.php
@@ -1,0 +1,40 @@
+<?php
+
+/**
+ * File containing the SystemInfoCollectorPass class.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzSupportToolsBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+class SystemInfoCollectorPass implements CompilerPassInterface
+{
+    /**
+     * Registers the SystemInfoCollector into the system info collector registry.
+     *
+     * @param \Symfony\Component\DependencyInjection\ContainerBuilder $container
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->has('support_tools.info_collector.registry')) {
+            return;
+        }
+
+        $infoCollectorsTagged = $container->findTaggedServiceIds('support_tools.info_collector');
+
+        $infoCollectors = [];
+        foreach ($infoCollectorsTagged as $id => $tags) {
+            foreach ($tags as $attributes) {
+                $infoCollectors[$attributes['alias']] = new Reference($id);
+            }
+        }
+
+        $infoCollectorRegistryDef = $container->findDefinition('support_tools.info_collector.registry');
+        $infoCollectorRegistryDef->setArguments([$infoCollectors]);
+    }
+}
--- a/Resources/config/services.yml
+++ b/Resources/config/services.yml
@@ -4,6 +4,7 @@ imports:
 parameters:
     support_tools.command.dump_info.class: EzSystems\EzSupportToolsBundle\Command\SystemInfoDumpCommand
     support_tools.info_collector.registry.class: EzSystems\EzSupportToolsBundle\SystemInfo\Registry\IdentifierBased
+    support_tools.info_collector.system_info.ezc.factory.class: EzSystems\EzSupportToolsBundle\SystemInfo\EzcSystemInfoFactory
     support_tools.info_collectors.composer.lock_file.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\JsonComposerLockSystemInfoCollector
     support_tools.info_collectors.database.doctrine.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\DoctrineDatabaseSystemInfoCollector
     support_tools.info_collectors.hardware.ezc.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\EzcHardwareSystemInfoCollector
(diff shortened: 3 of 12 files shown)


DependencyInjection/Compiler/SystemInfoCollectorPass.php
```diff
@@ -0,0 +1,37 @@
...
+        if (!$container->has('support_tools.info_collector.registry')) {
+            return;
+        }
+
+        $infoCollectorRegistry = $container->findDefinition('support_tools.info_collector.registry');
+        $infoCollectors = $container->findTaggedServiceIds('support_tools.info_collector');
+
+        foreach ($infoCollectors as $id => $tags) {
+            foreach ($tags as $attributes) {
+                $infoCollectorRegistry->addMethodCall('addToRegistry', [new Reference($id), $attributes['alias']]);
```
bdunogier: A performance improvement we try to apply: instead of adding N method calls, add one method call with N collectors. You could just make `addToRegistry` accept a hash of collectors, indexed by identifier.

SystemInfo/SystemInfoCollectorRegistry.php
```diff
@@ -0,0 +1,57 @@
...
+
+        return null;
+    }
+
+    /**
+     * Returns all registered SystemInfoCollectors.
+     *
+     * @return \EzSystems\EzSupportToolsBundle\SystemInfo\Collector\SystemInfoCollector[] Hash of SystemInfoCollectors, with service tag alias as keys.
+     */
+    public function getAllFromRegistry()
```
bdunogier: I'm not sure I'd expose that. What about `getKeys`, or `getIdentifiers` ?

SystemInfo/SystemInfoCollectorRegistry.php
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    /**
+     * Returns the SystemInfoCollector matching the argument.
+     *
+     * @param string $serviceTagAlias A service tag alias to match against. Example: composer
+     *
+     * @return \EzSystems\EzSupportToolsBundle\SystemInfo\Collector\SystemInfoCollector|null The SystemInfoCollector, or null if there is none.
+     */
+    public function getFromRegistry($serviceTagAlias)
```
bdunogier: I wouldn't repeat "registry" in the method's name

SystemInfo/SystemInfoCollectorRegistry.php
glye: To paraphrase you: "A performance improvement I try to apply: instead of adding N method calls, add one method call with N collectors" :P
The CLI command has a usecase for getting all collectors at once. But I'll remove it if you insist :)

SystemInfo/SystemInfoCollectorRegistry.php
glye: Good, with that perspective it is generic enough that I could make an interface in front and have a specific "ServiceTag" implementation of it, solving the extensibility concern. (Well, apart from the fact that not all implementations may need an "identifier" parameter as it may be included in the object. But they could just pass false then.)

As for naming I'm being consistent with the viewbuilder registry :) But just ´getFrom´ sounds bad. How about `getItem`, or `get`?

DependencyInjection/Compiler/SystemInfoCollectorPass.php
glye: Yes, sounds good.

glye:
@bdunogier Updated: Using an interface now, adding all items in one go as a hash, renaming, don't expose whole registry array.

Command/SystemInfoDumpCommand.php
```diff
@@ -39,11 +54,39 @@ protected function configure()
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $infoCollector = $this->getContainer()->get($input->getArgument('info-collector'));
+        $registry = $this->getContainer()->get('support_tools.info_collector.registry');
```
glye: I'll inject the registry instead.

glye:
Update: Made the dump command a service, inject the registry.

Command/SystemInfoDumpCommand.php
```diff
@@ -8,25 +8,55 @@
  */
 namespace EzSystems\EzSupportToolsBundle\Command;
 
+use EzSystems\EzSupportToolsBundle\SystemInfo\Collector\SystemInfoCollector;
```
andrerom: please find something else then "InfoCollector"

glye: Ok, but that should go into a followup methinks.

"Gatherer" or "Hoarder" should be safe :) http://www.thesaurus.com/browse/collect?s=t

glye:
Update: Made the View work with the registry

glye:
Update: Added tests.
Ready for review @bdunogier @andrerom

Command/SystemInfoDumpCommand.php
glye: Ref: https://jira.ez.no/browse/EZP-25647

glye:
Update: Lazy loading the zeta components factory, to optimise the registry. Verified that it works.

Command/SystemInfoDumpCommand.php
bdunogier: Any suggestion ?

glye: I like `SystemInfoGatherer` - it means pretty much the same, it should be unique, and with that we can keep the `SystemInfo` value object naming.

bdunogier:
> Update: Lazy loading the zeta components factory, to optimise the registry. Verified that it works.

I'd have kept this in a distinct commit.

SystemInfo/SystemInfoCollectorRegistry.php
```diff
@@ -0,0 +1,38 @@
...
+ * A registry of SystemInfoCollectors.
+ */
+interface SystemInfoCollectorRegistry
+{
+    /**
+     * Adds SystemInfoCollectors, each with an identifier string, to the registry.
+     *
+     * @param \EzSystems\EzSupportToolsBundle\SystemInfo\Collector\SystemInfoCollector[] $items A hash of SystemInfoCollectors, with identifier string as key.
+     */
+    public function addItems(array $items);
```
bdunogier: Thinking, not for immediate application: I think we should (overall) separate the "add" methods from the "get" ones. What matters from a consumer's perspective are the "get" ones, and they don't care how items are inputed. It also gives us the liberty of having registries that are NOT based on items added over a method.

glye:
> I'd have kept this in a distinct commit.

It is: https://github.com/ezsystems/ez-support-tools/pull/12/commits/40d0603f66d572dc7c3a5909088658948d4a8b7a
You mean separate PR?

SystemInfo/SystemInfoCollectorRegistry.php
glye: If so... that would mean removing the "adder" ( :snake: :grinning:) from the interface, right? Even if we do separate them like this it doesn't hurt that they have consistent naming, IMO.

SystemInfo/Registry/IdentifierBased.php
```diff
@@ -0,0 +1,56 @@
...
+     *
+     * @return \EzSystems\EzSupportToolsBundle\SystemInfo\Collector\SystemInfoCollector|null The SystemInfoCollector, or null if there is none.
+     */
+    public function getItem($identifier)
+    {
+        if (isset($this->registry[$identifier])) {
+            return $this->registry[$identifier];
+        }
+
+        return null;
```
bdunogier: I would make it throw an exception if the item doesn't exist. Do you think it is too much ? Would avoid the return value handling in the higher layers.

Command/SystemInfoDumpCommand.php
bdunogier: I understand.

Gotta admit that while reading the code, some things could have been done differently: `support_tools.info_collector.registry` could be named `support_tools.system_info.registry_collector`, or `support_tools.system_info.collector.registry` (e.g. always prefix "info" with "collector", separated by `_` to indicate that it is one word ("info_collector" not being a name in the domain of ez support tools).

SystemInfo/SystemInfoCollectorRegistry.php
bdunogier: >  it doesn't hurt that they have consistent naming, IMO.

Not sure what you mean exactly, but I'm all for consistency :-)

Command/SystemInfoDumpCommand.php
glye: Uh, that's somewhat unclear, but I agree that `support_tools.system_info.collector_registry` is better than `support_tools.info_collector.registry`.

Having `support_tools.system_info` before other parts is good. Both `collector_registry` and `collector.registry` works, `registry_collector` not so much.

SystemInfo/SystemInfoCollectorRegistry.php
glye: You commented on "registry" in the name of the getter, not for the setter. I removed it from both, to keep them consistent. Could be what you meant, of course :)

SystemInfo/Registry/IdentifierBased.php
glye: OK with me. We're throwing an exception in SystemInfoViewBuilder, might as well do it at the source of the problem instead.

glye:
Update: Throw NotFoundException in the registry, instead of downstream.

View/SystemInfoViewBuilder.php
```diff
@@ -47,14 +49,9 @@ public function buildView(array $parameters)
      * @param string $identifier A SystemInfo collector identifier (php, hardware...)
      *
      * @return \EzSystems\EzSupportToolsBundle\SystemInfo\Collector\SystemInfoCollector
-     * @throws \eZ\Publish\Core\Base\Exceptions\NotFoundException If no SystemInfoCollector exists with this identifier
      */
     private function getCollector($identifier)
```
glye: @bdunogier Do you want to keep this method, given that it doesn't do much anymore?

bdunogier:
>> I'd have kept this in a distinct commit.
> It is: 40d0603
> You mean separate PR?

No, separate commit. A !fixup commit is meant to be squashed onto the other ones, and is therefore not really distinct. You could just rebase this particular commit, and give it a distinctive commit message ("Made the zeta ezcSysInfo factory lazy". You may then also add extra info about why this commit was necessary in the second part of the commit message).

glye:
@bdunogier Ok, thought you meant distinct for the purpose of this review. Can do. Could you explain why you want that commit? Isn't it more accessible to have this info in the doc of the factory class, if we're going to be verbose about it?

Command/SystemInfoDumpCommand.php
bdunogier: `registry_collector` was a mistake. `collector_registry` was the intention, good catch :-)

bdunogier:
> Could you explain why you want that commit?

Because it is a distinct change, unrelated to the topic of that pull-request.

glye:
> unrelated to the topic of that pull-request.

Sorry, I don't get you. This PR adds new functionality requiring that collector constructors be lightweight. So how is the lazy-loading unrelated?

bdunogier:
Nevermind, then.

glye:
Update: Reorganised the PR commits to clarify changes. CC @bdunogier

glye:
> The fact that it highlights a performance issue in a component used by two of the collectors is not "part of this change".

That *became* a performance issue because of this change, it wasn't before. So to me it is part of the change. But I have reorganised it, in the interest of peace and harmony :)

Command/SystemInfoDumpCommand.php
andrerom: > Any suggestion ? Note that it is a SystemInfoCollector, not an InfoCollector.

Coming back to this _(sorry for late reply)_, I see that as well. I think I reacted since it was referred to during daly or something as infoCollector. As long as we refer to it by full name I'm fine with it.

bdunogier: Glad we've cleared it up.

The notes above about consistency in services & classes names must be taken care of anyway I think. We are indeed close from a legacy domain concept, and we must make sure there are no inconsistencies there.

glye: Yep, good to hear. I have updated https://jira.ez.no/browse/EZP-25647 accordingly.

glye:
AFAICT we're done here. Cautiously optimistic final review ping @bdunogier @andrerom

SystemInfo/SystemInfoCollectorRegistry.php
andrerom: Not really following the discussion here.

But another point; what about renaming it `setItems`, in real life you only need to set it once right? And if someone, even tests needs to append, then they can use `getItems` to append and then set. This will avoid the need to do array merge also.

glye: I'm fine with this. @bdunogier?

andrerom: Well I understood what @bdunogier implied here now, and I'll have to agree, this should just be set in ctor in one go, that would solve my remark above also, for test you'll typically init it in setup with what you want to test.

See eZ\Publish\Core\Repository\Helper\FieldTypeRegistry for example.

Resources/config/services.yml
```diff
@@ -9,32 +12,56 @@ parameters:
...
 
 services:
+    support_tools.command.dump_info:
+        class: %support_tools.command.dump_info.class%
+        arguments:
+            - @support_tools.info_collector.registry
+        tags:
+            - { name: console.command }
+
+    support_tools.info_collector.registry:
```
bdunogier: Naming: `support_tools.system_info.collector_registry` ?

Resources/config/services.yml
```diff
@@ -9,32 +12,56 @@ parameters:
...
             - @support_tools.info_collector.system_info.ezc
+        tags:
+            - { name: "support_tools.info_collector", alias: "hardware" }
 
     support_tools.info_collectors.php.ezc:
         class: %support_tools.info_collectors.php.ezc.class%
         arguments:
             - @support_tools.info_collector.system_info.ezc
+        tags:
+            - { name: "support_tools.info_collector", alias: "php" }
```
bdunogier: We also need to rename the service tag, btw. `support_tools.system_info_collector` ?

Also, we use `identifier` in the registry's code, not `alias`.

SystemInfo/SystemInfoCollectorRegistry.php
bdunogier: What bothers me a bit with using the ctor is that this aspect of the service is _not_ part of the interface... 

But it is also the simplest approach, so why not.

bdunogier:
Aaaaand +1, as long as we take care of naming consistency in a not too distant future.

glye:
Update: Removed registry `addItems()`, use constructor instead.

Resources/config/services.yml
glye: Might as well get it done.

glye:
Update: Services renaming
`support_tools.info_collectors` -> `support_tools.system_info.collector`
`support_tools.info_collector.registry` -> `support_tools.system_info.collector_registry`
`support_tools.info_collector.system_info.ezc` -> `support_tools.system_info.ezc`
`alias` -> `identifier`

glye:
Rebased on master after https://github.com/ezsystems/ez-support-tools/pull/14

Command/SystemInfoDumpCommand.php
```diff
@@ -8,25 +8,55 @@
...
+
+    /**
      * Define command and input options.
      */
     protected function configure()
     {
         $this
             ->setName('ez-support-tools:dump-info')
-            ->setDescription('Collects system information and dumps it.')
+            ->setDescription(<<<'EOD'
```
bdunogier: The description should be one line, or we get this when running `app/console`:

```
 ez-support-tools
  ez-support-tools:dump-info            Collects system information and dumps it.

By default it dumps information from all available information collectors.
You can specify one or more collectors as arguments, e.g. 'php database hardware'.
To get a list if available collectors, use '--list-info-collectors'
 ezplatform
```

This should do:
```
            ->setDescription('Collects system information and dumps it.')
            ->setHelp(<<<EOH
By default it dumps information from all available information collectors.
You can specify one or more collectors as arguments, e.g. 'php database hardware'.
To get a list if available collectors, use '--list-info-collectors'
EOH
            )
```

andrerom:
+1 with suggested changes for better help text

Command/SystemInfoDumpCommand.php
glye: Thanks, that's useful. Neither console tutorials nor classdoc mention it afaict.

glye:
Update: Fixed setDescription/setHelp in the command.

glye:
Failure: `Argument #3 (No Value) of PHPUnit_TextUI_ResultPrinter::__construct() must be a value from "never", "auto" or "always"`
Fails on PHP 5.x, works on PHP7. What's wrong here - wrong phpunit version installed on travis?

glye:
Travis/phpunit should be good now. Also squashed the last two fixups.

PR #13, opened 2016-04-05
Added default system_info_view settings

bdunogier:
Avoids this error:

```
  [eZ\Publish\Core\MVC\Exception\ParameterNotFoundException]
  Parameter 'system_info_view' with namespace '' could not be found.
```


glye:
+1

PR #15, opened 2016-04-11
EZP-25653: Allow use of eZSupportTools in PlatformUI

glye:
> Part of https://jira.ez.no/browse/EZP-25653
> Status: Ready for review

Removed default_settings.yml as it conflicts with the same in platformui (it resets the collectors).
Ref https://github.com/ezsystems/PlatformUIBundle/pull/542

### DIFF
--- a/DependencyInjection/EzSystemsEzSupportToolsExtension.php
+++ b/DependencyInjection/EzSystemsEzSupportToolsExtension.php
@@ -22,6 +22,5 @@ class EzSystemsEzSupportToolsExtension extends Extension
     {
         $loader = new Loader\YamlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
         $loader->load('services.yml');
-        $loader->load('default_settings.yml');
     }
 }
--- a/Resources/config/default_settings.yml
+++ b/Resources/config/default_settings.yml
@@ -1,2 +1,0 @@
-parameters:
-    ezsettings.default.system_info_view: {}


bdunogier:
The title can't be right: ez-support-tools doesn't know about PlatformUI ;-)

Jokes aside, on its own, ezsystems/ez-support-tools will break with this patch: it will try to use `$system_info_view$` as a service argument. We can't rely on platform-ui-bundle declaring it.

But it is true that this will break platform-ui. I see three ways around this:
1. load the support tools bundle *before* PlatformUI, so that the latter overrides the empty setting
2. add a compiler pass to support tools that declares this setting if it has not been declared
3. implement semantic config (I have a skeleton for this)

The commit must be changed to say what it does (`Removed default_settings.yml`). So would the PR, but it's not as important if it ends up containing more commits.

In any case, I can't be +1 as it is (break).

glye:
Thanks for the help. 1 or 2 seem the faster ways to get to a usable solution.

> The title can't be right: ez-support-tools doesn't know about PlatformUI ;-) Jokes aside(...)

I don't get the joke, the title isn't assuming ez-support-tools knows about platformui...?

bdunogier:
Damn, probably french / my type of humour :-)

Anyway, the end of my comment should explain what I mean in reality.

bdunogier:
Solution is probably the easiest, since you will have to create an ezplatform PR (see platformUI comment). You can set the order here, and it should work as expected.

andrerom:
Title fixed.

glye:
Closing, not needed.

PR #17, opened 2016-08-04
EZP-26153: ez-support-tools database collector is not siteaccess aware

iosolinf:
In order to select active database, both SiteAccess and doctrine.dbal database list were injected into service

### DIFF
--- a/Resources/config/services.yml
+++ b/Resources/config/services.yml
@@ -42,10 +42,11 @@ services:
 
     support_tools.system_info.collector.database.doctrine:
         class: %support_tools.system_info.collector.database.doctrine.class%
-        arguments:
-            - @database_connection
+        calls:
+            - [setSiteAccess, [@ezpublish.siteaccess]]
+            - [setDatabaseList, [@doctrine]]
         tags:
-            - { name: "support_tools.system_info.collector", identifier: "database" }
+             - { name: "support_tools.system_info.collector", identifier: "database" }
 
     support_tools.system_info.collector.hardware.ezc:
         class: %support_tools.system_info.collector.hardware.ezc.class%
--- a/SystemInfo/Collector/DoctrineDatabaseSystemInfoCollector.php
+++ b/SystemInfo/Collector/DoctrineDatabaseSystemInfoCollector.php
@@ -9,6 +9,8 @@
 namespace EzSystems\EzSupportToolsBundle\SystemInfo\Collector;
 
 use Doctrine\DBAL\Connection;
+use Doctrine\Bundle\DoctrineBundle\Registry as DoctrineRegistry;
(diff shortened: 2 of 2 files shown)


Resources/config/services.yml
```diff
@@ -42,10 +42,11 @@ services:
...
     support_tools.system_info.collector.database.doctrine:
         class: %support_tools.system_info.collector.database.doctrine.class%
-        arguments:
-            - @database_connection
+        calls:
+            - [setSiteAccess, [@ezpublish.siteaccess]]
+            - [setDatabaseList, [@doctrine]]
         tags:
-            - { name: "support_tools.system_info.collector", identifier: "database" }
+             - { name: "support_tools.system_info.collector", identifier: "database" }
```
joaoinacio: CS: extra space?

SystemInfo/Collector/DoctrineDatabaseSystemInfoCollector.php
```diff
@@ -39,11 +46,24 @@ public function __construct(Connection $db)
      */
     public function collect()
     {
+        $this->connection = $this->connectionList[$this->siteAccess->name];
```
joaoinacio: IMO it doesn't seem necessary to use a `connection` property if it's not used elsewhere, could just as well use a simple variable.

iosolinf: :) That would had been too easy, I'm afraid...
https://doc.ez.no/display/DEVELOPER/SiteAccess#SiteAccess-Limitations includes the following:
Injecting dynamic settings in request listeners is not recommended, as it won't be resolved with the correct scope (request listeners are instantiated before SiteAccess match). Workaround is to inject the ConfigResolver instead, and resolving the setting in your onKernelRequest method (or equivalent).

In constructor the siteaccess is always undefined.

andrerom: What about using the ConfigResolver then? :)

joaoinacio:
After looking at this I believe the simplest fix would probably be to fetch the db connection from ezp (which uses the api `StorageConnectionFactory`) and tag the service as lazy:
```yml
    support_tools.system_info.collector.database.doctrine:
        class: %support_tools.system_info.collector.database.doctrine.class%
        arguments:
            - @ezpublish.persistence.connection
        lazy: true
        tags:
            - { name: "support_tools.system_info.collector", identifier: "database" }
```

iosolinf:
1000 thousand times better, @joaoinacio big +1

andrerom:
+1

iosolinf:
ping @glye , @bdunogier Anyone else for a +1 ?

glye:
+1 apart from that you should double quote `@params` (and `%params%`).

PR #18, opened 2016-08-05
EZP-26549: Output pretty JSON in dump-info command

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-26549

Current output format uses multiple calls to `var_export`, which has two problems:
- It does not identify the "name" of the information category/collector.
- It uses a "not-that-common" format that is not as easily parsed/processed by some tools as it could be:

```
array (
  'type' => 'mysql',
  'name' => 'ezenterprise14',
  'host' => 'localhost',
  'username' => 'root',
)
array (
  'cpuType' => 'Intel(R) Core(TM) i5-2410M CPU @ 2.30GHz',
  'cpuSpeed' => 2294.7860000000001,
  'cpuCount' => 2,
  'memorySize' => 1577353216,
)
array (
  'version' => '5.6.20-0+deb8u1',
  'acceleratorEnabled' => true,
  'acceleratorName' => 'Zend OPcache',
  'acceleratorURL' => 'http://www.php.net/opcache',
  'acceleratorVersion' => '7.0.6-devFE',
)
```

This changes the output format to a single json object, with tries to solve both problems:

``` json
{
    "database": {
        "type": "mysql",
        "name": "ezenterprise14",
        "host": "localhost",
        "username": "root"
    },
    "hardware": {
        "cpuType": "Intel(R) Core(TM) i5-2410M CPU @ 2.30GHz",
        "cpuSpeed": 2294.786,
        "cpuCount": 2,
        "memorySize": 1577353216
    },
    "php": {
        "version": "5.6.20-0+deb8u1",
        "acceleratorEnabled": true,
        "acceleratorName": "Zend OPcache",
        "acceleratorURL": "http:\/\/www.php.net\/opcache",
        "acceleratorVersion": "7.0.6-devFE"
    }
}
```

### DIFF
--- a/Command/SystemInfoDumpCommand.php
+++ b/Command/SystemInfoDumpCommand.php
@@ -73,36 +73,17 @@ EOD
             foreach ($this->registry->getIdentifiers() as $identifier) {
                 $output->writeln("  $identifier", true);
             }
-        } else if ($identifiers = $input->getArgument('info-collectors')) {
-            foreach ($identifiers as $identifier) {
-                $this->outputInfo(
-                    $this->registry->getItem($identifier),
-                    $output
-                );
-            }
+        } else if ($input->getArgument('info-collectors')) {
+            $identifiers = $input->getArgument('info-collectors');
         } else {
-            foreach ($this->registry->getIdentifiers() as $identifier) {
-                $this->outputInfo($this->registry->getItem($identifier), $output);
-            }
-        }
-    }
-
-    /**
-     * Output info collected by the given collector.
-     *
-     * @param $infoCollector SystemInfoCollector
-     * @param $output OutputInterface
-     */
-    private function outputInfo(SystemInfoCollector $infoCollector, OutputInterface $output)
-    {
-        $infoValue = $infoCollector->collect();
-
-        $outputArray = [];
-        // attributes() is deprecated, and getProperties() is protected. TODO add a toArray() or similar.
-        foreach ($infoValue->attributes() as $property) {
-            $outputArray[$property] = $infoValue->$property;
+            $identifiers = $this->registry->getIdentifiers();
         }
 
-        $output->writeln(var_export($outputArray, true));
+        $collectedInfo = [];
+        foreach ($identifiers as $identifier) {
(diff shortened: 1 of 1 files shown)


iosolinf:
+1

bdunogier:
Looks good to me. @glye ?

Command/SystemInfoDumpCommand.php
```diff
@@ -73,36 +73,17 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-    private function outputInfo(SystemInfoCollector $infoCollector, OutputInterface $output)
-    {
-        $infoValue = $infoCollector->collect();
 
-        $outputArray = [];
-        // attributes() is deprecated, and getProperties() is protected. TODO add a toArray() or similar.
-        foreach ($infoValue->attributes() as $property) {
-            $outputArray[$property] = $infoValue->$property;
+        $collectedInfo = [];
+        foreach ($identifiers as $identifier) {
```
glye: Don't you get an "undefined variable" warning here when using the `list-info-collectors` option?

joaoinacio: Good point, not certain if that was tested but probably not; will fix.

andrerom:
> @bdunogier: any idea what other formats would be useful? xml ? perhaps something for a separate enhancement PR

it's a bit for separate story, but if you see a easy way to use tagged services for this then you could add the infrstructte for it already and make json default, the only one for now.

bdunogier:
@joaoinacio actually yes, make that an improvement, and change the default format to json. The array thing was really to have some kind of output.

XML sounds like a good option to have. Make sure to specify in the JIRA improvement that it should be easy to add/define new formats.

bdunogier:
P.S. Was a JIRA issue created ? The description says that there isn't :)

andrerom:
@bdunogier seems there is an issue now ;)

joaoinacio:
> it's a bit for separate story, but if you see a easy way to use tagged services for this then you could add the infrstructte for it already and make json default, the only one for now.

Had a quick look, seems a bit more work (tagged services, compiler pass, registry, tests, etc), probably worth it do a separate PR.

PR #20, opened 2016-10-07
EZP-26391: ez-support-tools factory called on every request

glye:
> Fixes https://jira.ez.no/browse/EZP-26391
> Ready for review

It's not the factory that's lazy, but the service it builds.


andrerom:
+1

PR #21, opened 2016-10-18
EZP-26391: ez-support-tools factory called on every request

glye:
> Fixes https://jira.ez.no/browse/EZP-26391
> Part 1: https://github.com/ezsystems/ez-support-tools/pull/20
> Status: QA

Lazy loading was failing because ezcSystemInfo uses magic properties, while the proxy manager looks for public properties. This fixes the issue by wrapping ezcSystemInfo and exposing the magic properties as public ones.

Manual test: Verified that both the web interface and the command work as expected, and that the factory build method is only called when needed.


bdunogier:
Looks okay.

What bothers me a tiny bit is that there _is_ code in the class generated for the lazy proxy to handle `__get()`, and I'm not really sure why it doesn't work in dev, but does in prod. Probably missing something, but it's not really code that is meant to be read.

glye:
Fixing failing tests.

glye:
Ping @andrerom @yannickroger - this fixes the problem QA found in https://github.com/ezsystems/ez-support-tools/pull/20

andrerom:
Did you by any chance look at if we can use Proxy manager 2.x? It might solve the issue.
https://github.com/Ocramius/ProxyManager/releases

bdunogier:
I haven't yet. Let's create a story for this.

andrerom:
it requires php 7, ignore..

bdunogier:
Frak :(

glye:
Ack. Time may be too short to get this fix in by release, I'll un-lazy the service later today if we don't make it, and re-do the PR.

SystemInfo/EzcSystemInfoWrapper.php
```diff
@@ -0,0 +1,72 @@
...
+
+    /** @var \ezcSystemInfoAccelerator */
+    public $phpAccelerator;
+
+    /** @var bool */
+    public $isShellExecution;
+
+    public function __construct()
+    {
+        $ezcSystemInfo = ezcSystemInfo::getInstance();
```
andrerom: Could you please also take care about the exception while on it, as the upstream pull is taking a long time, ref:
```php
        try {
            $ezcSystemInfo = ezcSystemInfo::getInstance();
        } catch(ezcSystemInfoReaderCantScanOSException $e) {
            // Leave properties as null: https://github.com/zetacomponents/SystemInformation/pull/9
            return;
        }
```

glye:
Making the service non-lazy for now: https://github.com/ezsystems/ez-support-tools/pull/22
but I'll resolve this.

SystemInfo/EzcSystemInfoFactory.php
```diff
@@ -8,15 +8,13 @@
...
-
 /**
  * Factory for zetacomponents/sysinfo. Used here to simplify testing.
  */
 class EzcSystemInfoFactory
 {
     public static function buildEzcSystemInfo()
     {
-        return ezcSystemInfo::getInstance();
+        return new EzcSystemInfoWrapper();
```
glye: Affirmative.

glye:
Update: Removed the redundant factory, and catch `ezcSystemInfoReaderCantScanOSException`.

andrerom:
Nice, wdt @bdunogier ? Seems like the way to go imoh over #22, as this will also fix issues people have with blocked usage when ezcSystemInfoReader is checked and can't read system info, #22 won't afaik.

glye:
> #22 will have that issue afaik, as it's not lazy and thus will be executed all the time

True.

bdunogier:
I indeed prefer this one too. Could QA test it ? I'll have another look and hopefully approve it.

SystemInfo/EzcSystemInfoWrapper.php
```diff
@@ -0,0 +1,78 @@
...
+    public function __construct()
+    {
+        try {
+            $ezcSystemInfo = ezcSystemInfo::getInstance();
+        } catch(ezcSystemInfoReaderCantScanOSException $e) {
+            // Leave properties as null: https://github.com/zetacomponents/SystemInformation/pull/9
+            return;
+        }
+
+        $this->osType = $ezcSystemInfo->osType;
```
bdunogier: You could probably shorten that to some kind of loop:

```
foreach (get_object_vars($this) as $var) {
  $this->$var = $ezcSystemInfo->$var;
}
```

bdunogier:
Let's send that to QA asap. The issue is annoying enough, and it'd be a shame not to fix it. Worst thing that happen, we'll do another support tools RC :)

SystemInfo/EzcSystemInfoWrapper.php
andrerom: well, it also makes it less readable, so if QA approves it like this lets then maybe skip changing this part if that is ok.

bdunogier: No strong feelings :)

glye:
Update: Shortened wrapper property assignment.

pcardiga:
PR verified by QA

glye:
Thanks! Just waiting for Travis then...

PR #23, opened 2016-11-04
EZP-26549: Output pretty JSON in dump-info command

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-26549

Implementation of JSON systeminfo output format using tagged services.

Replaces https://github.com/ezsystems/ez-support-tools/pull/18

### DIFF
--- a/Command/SystemInfoDumpCommand.php
+++ b/Command/SystemInfoDumpCommand.php
@@ -24,7 +24,7 @@ class SystemInfoDumpCommand extends ContainerAwareCommand
      *
      * @var \EzSystems\EzSupportToolsBundle\SystemInfo\SystemInfoCollectorRegistry
      */
-    private $registry;
+    private $systemInfoCollectorRegistry;
 
     /**
      * Output format registry.
@@ -33,9 +33,9 @@ class SystemInfoDumpCommand extends ContainerAwareCommand
      */
     private $outputFormatRegistry;
 
-    public function __construct(SystemInfoCollectorRegistry $registry, OutputFormatRegistry $outputFormatRegistry)
+    public function __construct(SystemInfoCollectorRegistry $systemInfoCollectorRegistry, OutputFormatRegistry $outputFormatRegistry)
     {
-        $this->registry = $registry;
+        $this->systemInfoCollectorRegistry = $systemInfoCollectorRegistry;
         $this->outputFormatRegistry = $outputFormatRegistry;
 
         parent::__construct();
@@ -85,7 +85,7 @@ EOD
      {
          if ($input->getOption('list-info-collectors')) {
              $output->writeln('Available info collectors:', true);
-             foreach ($this->registry->getIdentifiers() as $identifier) {
+             foreach ($this->systemInfoCollectorRegistry->getIdentifiers() as $identifier) {
                  $output->writeln("  $identifier", true);
              }
              return;
@@ -98,13 +98,13 @@ EOD
          if ($input->getArgument('info-collectors')) {
              $identifiers = $input->getArgument('info-collectors');
          } else {
-             $identifiers = $this->registry->getIdentifiers();
+             $identifiers = $this->systemInfoCollectorRegistry->getIdentifiers();
          }
 
          // Collect info for the given identifiers.
          $collectedInfoArray = [];
          foreach ($identifiers as $identifier) {
-             $collectedInfoArray[$identifier] = $this->registry->getItem($identifier)->collect();
+             $collectedInfoArray[$identifier] = $this->systemInfoCollectorRegistry->getItem($identifier)->collect();
          }
 
          $output->writeln(
--- a/Command/SystemInfoDumpCommand.php
+++ b/Command/SystemInfoDumpCommand.php
@@ -63,9 +63,10 @@ EOD
             )
             ->addOption(
                 'format',
-                'json',
+                'f',
                 InputOption::VALUE_OPTIONAL,
-                'Output format (currently only JSON)'
+                'Output format (currently only JSON)',
+                'json'
             )
             ->addArgument(
                 'info-collectors',
--- a/Command/SystemInfoDumpCommand.php
+++ b/Command/SystemInfoDumpCommand.php
@@ -10,6 +10,7 @@ namespace EzSystems\EzSupportToolsBundle\Command;
 
 use EzSystems\EzSupportToolsBundle\SystemInfo\Collector\SystemInfoCollector;
 use EzSystems\EzSupportToolsBundle\SystemInfo\SystemInfoCollectorRegistry;
+use EzSystems\EzSupportToolsBundle\SystemInfo\OutputFormatRegistry;
 use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
 use Symfony\Component\Console\Input\InputArgument;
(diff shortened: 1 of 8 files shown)


Command/SystemInfoDumpCommand.php
```diff
@@ -66,43 +81,34 @@ protected function configure()
...
-        }
-    }
+     protected function execute(InputInterface $input, OutputInterface $output)
+     {
+         if ($input->getOption('list-info-collectors')) {
+             $output->writeln('Available info collectors:', true);
+             foreach ($this->registry->getIdentifiers() as $identifier) {
+                 $output->writeln("  $identifier", true);
+             }
+         } else {
```
andrerom: you can do a return after foreach above and get rid of the else indentation for the remaining code

joaoinacio:
ping @bdunogier @glye

DependencyInjection/Compiler/SystemInfoOutputFormatPass.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+/**
+ * File containing the SystemInfoCollectorPass class.
```
glye: Copypasta error? :)

DependencyInjection/Compiler/SystemInfoOutputFormatPass.php
```diff
@@ -0,0 +1,40 @@
...
+namespace EzSystems\EzSupportToolsBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+class SystemInfoOutputFormatPass implements CompilerPassInterface
+{
+    /**
+     * Registers the SystemInfoCollector into the system info collector registry.
```
glye: OutputFormat, not SystemInfoCollector

DependencyInjection/Compiler/SystemInfoOutputFormatPass.php
```diff
@@ -0,0 +1,40 @@
...
+     *
+     * @param \Symfony\Component\DependencyInjection\ContainerBuilder $container
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->has('support_tools.system_info.output_registry')) {
+            return;
+        }
+
+        $infoCollectorsTagged = $container->findTaggedServiceIds('support_tools.system_info.output_format');
```
glye: outputFormats, not infoCollectors

DependencyInjection/Compiler/SystemInfoOutputFormatPass.php
```diff
@@ -0,0 +1,40 @@
...
+
+        $infoCollectors = [];
+        foreach ($infoCollectorsTagged as $id => $tags) {
+            foreach ($tags as $attributes) {
+                $infoCollectors[$attributes['identifier']] = new Reference($id);
+            }
+        }
+
+        $infoCollectorRegistryDef = $container->findDefinition('support_tools.system_info.output_registry');
```
glye: outputFormat, not infoCollector

SystemInfo/OutputFormat.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+/**
+ * File containing the Output class.
```
glye: OutputFormat

SystemInfo/OutputFormat/JsonOutputFormat.php
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzSupportToolsBundle\SystemInfo\OutputFormat;
+
+use EzSystems\EzSupportToolsBundle\SystemInfo\OutputFormat as SystemInfoOutputFormat;
+
+/**
+ * Collects hardware system information using zetacomponents/sysinfo.
```
glye: copypasta...

SystemInfo/OutputFormatRegistry.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+/**
+ * File containing the SystemInfoCollectorRegistry interface.
```
glye: OutputFormatRegistry
class, not interface

joaoinacio:
> See comments. Also, when we have more than one output format we'll need an option for listing the available formats, but that's not strictly needed now.

Thanks. should be easy to return directly from the registry (same as collectors). will fix c/p names etc.

Resources/config/services.yml
```diff
@@ -64,3 +69,10 @@ services:
             - "%kernel.bundles%"
         tags:
             - { name: "support_tools.system_info.collector", identifier: "symfony_kernel" }
+
+    # SystemInfoOutputFormats
+
+    support_tools.system_info.output_format.json:
+        class: "%support_tools.system_info.output_format.json.class%"
+        tags:
+            - { name: "support_tools.system_info.output_format", identifier: "json" }
```
andrerom: The [convention is to use ´alias´](http://symfony.com/doc/current/service_container/tags.html#adding-additional-attributes-on-tags), it's not a strong convention so I'll leave it up to @bdunogier if we should follow upstream or stay consistent with other tags here.

joaoinacio:
For adding an `XML` format a small issue is that - unlike `JSON` - it does **not** allow for forward slashes in element names (f.e. for composer packages), even escaped.

```json
{
    "composer": {
        "packages": {
            "doctrine\/annotations": {
                "name": "doctrine\/annotations",
```

```xml
<?xml version="1.0"?>
    <info>
        <composer>
            <packages>
                <doctrine\/annotations>
                    <name>
```
Even though SimpleXMLElement does not complain about this, passing it to a DOMDocument (for pretty print) results in an exception.
Any suggestions?

glye:
Replace them with underscore or period? Hyphen is also allowed, but is used by packages, I think.

Resources/config/services.yml
joaoinacio: Sounds good. Note the concrete implementation of InfoCollectorRegistry interface is "IdentifierBased" so in case the tag changes a minor refactoring/renaming would be needed for consistency (also for vars etc), but not a big deal imo.

joaoinacio:
> Replace them with underscore or period? Hyphen is also allowed, but is used by packages, I think.

Thanks, that was my initial thought as well.
Will submit a separate PR once this is merged in order to keep it simple.

bdunogier:
About the XML, why don't you just take a slightly different approach ?

```xml
<?xml version="1.0"?>
    <info>
        <composer>
            <packages>
              <package name="ezsystems/ezpublish-kernel" version="" />
              <package name="ezsystems/ezpublish-kernel" version="" />
            </packages>
```

Or am I missing something obvious ?

joaoinacio:
> About the XML, why don't you just take a slightly different approach ?
> ...
> Or am I missing something obvious ?

That seems simple but the problem is that the output formatter would require knowledge about the structure of the provided data, which is currently not the case.
If the data changes, or new info collectors are added, it could require an update to the formatter.

Perhaps there is a middle ground to the above, some known cases/ that require special handling? 
Good topic for discussion on incoming PR once this is merged.

PR #24, opened 2017-10-29
[Composer] Allowed ezpublish-kernel ^7.0

alongosz:
This PR allows `ezpublish-kernel:^7.0@dev` for `ez-support-tools`.

Reason: Right now we have to make workaround to install `ezplatform:2.0` meta-repository dependencies by setting requirement for `ezpublish-kernel` in the form of `7.0.x-dev as 6.99.x-dev` instead of simply `^7.0@dev`. This change solves that partially (see also: ezsystems/ezplatform-http-cache#16).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         }
     ],
     "require": {
-        "ezsystems/ezpublish-kernel": "^6.2",
+        "ezsystems/ezpublish-kernel": "^6.2 || ^7.0",
         "ocramius/proxy-manager": "~1.0|~2.0",
         "symfony/proxy-manager-bridge": "*",
         "zetacomponents/system-information": "^1.1"


andrerom:
+1, but you probably dont need to set dev, as usual that will be set in meta

composer.json
```diff
@@ -10,7 +10,7 @@
         }
     ],
     "require": {
-        "ezsystems/ezpublish-kernel": "^6.2",
+        "ezsystems/ezpublish-kernel": "^6.2 || ^7.0",
```
alongosz: > Inconsistent space usage compared to the line below, but I don't know which is correct..

It seems none :) This probably is due to some changes in composer.json format. Right now [doc mentions only double pipe `||` surrounded by spaces](https://getcomposer.org/doc/articles/versions.md#version-range), so I aligned this with it.
Thanks for pointing this out @glye

alongosz:
@glye @andrerom removed `@dev` and updated format of version constraint (see [comment](#discussion_r147733408))

PR #26, opened 2017-12-14
Fix undefined index notice when using path repos

emodric:
When using Composer path repositories, `time` and `source` are not present in package data, which crashes rendering.


andrerom:
@emodric Composer path repositories?

andrerom:
Ok: https://getcomposer.org/doc/05-repositories.md#path

andrerom:
NOTE: This might break UI now, so could also set some default values here if possible instead of setting null if that turns out to be the case.

emodric:
Thanks! Let me know if needed!

P.S. Tag would be appreciated :)

andrerom:
v0.2.1-rc1 tagged

PR #29, opened 2019-01-16
[WIP] Webpack Encore

Nattfarinn:
Follow up to: https://github.com/ezsystems/ezplatform/pull/345

Looks for webpack encore configuration per bundle in `Resources/encore/ez.config.js` directory.

Dumps list (JSON format) of path to files to `var/encore/ez.config.js` to be used by main feature.

Makes this part of main feature obsolete:

> It goes thru all bundles in vendor/ezsystems and looking for ez.webpack.config.js if found it register the config.

### DIFF
--- a/DependencyInjection/EzSystemsEzSupportToolsExtension.php
+++ b/DependencyInjection/EzSystemsEzSupportToolsExtension.php
@@ -11,10 +11,14 @@ namespace EzSystems\EzSupportToolsBundle\DependencyInjection;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Loader;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Finder\Finder;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 
 class EzSystemsEzSupportToolsExtension extends Extension
 {
+    const EZ_ENCORE_CONFIG_NAME = 'ez.config.js';
+
     /**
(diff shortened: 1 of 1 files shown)


andrerom:
Is there any particular reason this logic is placed in support-tools? _(assuming it's generic code needed for handling assets across all bundles)_

Nattfarinn:
@andrerom , only reason is that it does not fit anywhere right now. It was @lserwatka to put it here until we find better place (yet I need to put it somewhere so it can be used by @dew326 in his WIP).

It would be best to have something like `ezplatform-deploy-bundle` repository or something similar for such tools, but this is the only deploy tool I am aware of and it's just an Extension method, so not sure if another bundle is a way to go right now.

dew326:
@andrerom I'm not sure if we want to drop asset-bundle. 
Yes, now we could install our assets in the meta repo but this will be hard to maintain as we have 4 of them.

dew326:
When @Nattfarinn suggests me we can use asset-bundle for this I thought this is a good idea, but now when you pointed out that for each fix we will need to do a release I'm less enthusiastic.
This is a small piece of code and I do not assume we will have to fix something there but still.

lserwatka:
@dew326 it is not a big deal to have a maintenance releases there.

Nattfarinn:
Moved to: https://github.com/ezsystems/ezplatform-core

Ref. https://github.com/ezsystems/ezplatform-core/tree/master/src/EzPlatformEncoreBundle/bundle

lserwatka:
It is first step towards kernel rebranding, kernel split, and container for some core components.

PR #30, opened 2019-03-01
[Travis] Fix tests, move to newer version of phpunit

andrerom:
Some cleanup here too in order to align with other repos.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ before_script:
 
 script:
     - php vendor/bin/phpunit
-    - if [ "$CHECK_CS" = "true" ]; then bin/php-cs-fixer fix -v --dry-run --diff --show-progress=estimating; fi
+    - if [ "$CHECK_CS" = "true" ]; then vendor/bin/php-cs-fixer fix -v --dry-run --diff --show-progress=estimating; fi
 
 notifications:
     email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ matrix:
         - php: 7.2
         - php: 7.3
 
-# test only master (+ Pull requests)
+# test only on matching branches (+ Pull requests)
 branches:
     only:
         - master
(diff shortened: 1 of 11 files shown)


Tests/SystemInfo/Collector/ConfigurationSymfonyKernelSystemInfoCollectorTest.php
```diff
@@ -10,7 +10,7 @@
 
 use EzSystems\EzSupportToolsBundle\SystemInfo\Collector\ConfigurationSymfonyKernelSystemInfoCollector;
 use EzSystems\EzSupportToolsBundle\SystemInfo\Value\SymfonyKernelSystemInfo;
-use PHPUnit_Framework_TestCase;
+use PHPUnit\Framework\TestCase as PHPUnit_Framework_TestCase;
```
alongosz: Wouldn't it be simpler to change `extends PHPUnit_Framework_TestCase` into `extends TestCase`? Just one line more ;)

andrerom: sure nitty :)

andrerom: Fixed in e63eadc
Also bumped php-cs-fixer in 8438f57 so runs on PHP 7.3 works

.travis.yml
```diff
@@ -1,29 +1,39 @@
...
+    fast_finish: true
     include:
         - php: 5.6
         - php: 7.0
         - php: 7.1
           env: CHECK_CS=true
+        - php: 7.2
+        - php: 7.3
 
 # test only master (+ Pull requests)
```
glye: Not "only master" anymore given change below, right?

andrerom: indeed, fixed in 
a948ba1

PR #28, opened 2019-01-02
Basic version & system info on Editorial Dashboard

andrerom:
From support/services/community side we are in need of people being able to:
- known what version they are on _(at least minor, patch version is follow up in M2 by engineering)_
- know if something is wrongly setup or expired

The work here aims to do small first step, that can be shipped in a 0.2.x release of ez-support-tools so all existing 2.x installs gets access to this retroactively and gets info when version is out of date.

This is places on editorial dashboard for a few reasons:
- Minimal indication should be here to make sure project owners becomes aware of larger issues
    - Once we can place info in admin/systeminfo section we can make this link there instead.
- No way to extend tabs in SystemInfo section without having to do changes in Admin UI
    - So a follow up milestone (M2) for 2.5+ here for engineering 

Depends on the following:
- ~https://github.com/ezsystems/ezplatform-admin-ui/pull/783~ _For 2.4+ so this shows up above workflow block_
- https://github.com/ezsystems/developer-documentation/pull/494  _Doc coverage for how to move from TTL to BUL as this will make people aware they are running on wrong license_

### Screenshoot

platform_dashboard_trial_view

PR #31, opened 2019-03-04
Dashboard 1.13 support

ViniTou:
This PR fix adminUi dependency for older versions of eZPlatform.

Relevant:
https://github.com/ezsystems/ez-support-tools/pull/28


Resources/config/services.yml
```diff
@@ -34,6 +49,15 @@ services:
 
     # SystemInfoCollectors
 
+    support_tools.system_info.collector.system.ez:
+        class: "%support_tools.system_info.collector.system.ez.class%"
+        arguments:
+            - "@support_tools.system_info.collector.composer.lock_file"
+            - "%kernel.debug%"
+        # Can't tag this before v0.3 (2.5?) as it will blow up in admin UI for missing templates there
```
andrerom: 0.3 => 1.0

andrerom:
NOTE: Failure is on CS checks so it's relevant.

andrerom:
Todo for this backport: Remove Trial distinction.

Given there is bug in current builds which means all 2.x ee installs probably runs with TTL license packages. As 2.0-2.3 is End of Life, and 2.5 is End of Maintenance, it's better to focus fix on _that_ for 2.5.

andrerom:
I think we by now end up skipping this, ok with you @ViniTou ?

PR #35, opened 2019-04-26
Use correct key for update url

mnocon:
The URL is defined as `update`in https://github.com/ezsystems/ez-support-tools/blob/1.0/Resources/config/services.yml#L27

This problem occurred on current master in https://github.com/ezsystems/ezplatform/pull/397 in tests, when browsing the Dashboard:

https://res.cloudinary.com/ezplatformtravis/image/upload/v1556274925/screenshots/5cc2deed9682e714913640-vendor_ezsystems_ezplatform-admin-ui_features_objectstates_feature_11_zzymim.png


mnocon:
@lserwatka this one can be merged

PR #36, opened 2019-05-09
EZP-30402: Make Dashboard code more solid when it comes to detecting trial or not

andrerom:
Adds logic to also read info from composer.json to detect if packages are pulled from ttl channel
or not.

Besides the current issues with updates.ez.no telling composer the packages are ttl when they are not (diff between packages.json and composer.json). This also solves issues when people switch from TTL to BUL, and install won't show they have switched until all TTL packages actually have a new release to pull in.

### DIFF
--- a/Resources/config/services.yml
+++ b/Resources/config/services.yml
@@ -55,13 +55,14 @@ services:
             - "@support_tools.system_info.collector.composer.lock_file"
             - "%kernel.debug%"
         # Can't tag this before v0.3 (2.5?) as it will blow up in admin UI for missing templates there
-        # And it does not look like there is anway to add it from this package, so maybe it needs to be made extensible(?)
+        # And it does not look like there is anyway to add it from this package, so maybe it needs to be made extensible(?)
         #tags: [{ name: "support_tools.system_info.collector", identifier: "ez" }]
 
     support_tools.system_info.collector.composer.lock_file:
         class: "%support_tools.system_info.collector.composer.lock_file.class%"
         arguments:
             - "%kernel.root_dir%/../composer.lock"
+            - "%kernel.root_dir%/../composer.json"
         tags:
             - { name: "support_tools.system_info.collector", identifier: "composer" }
 
--- a/Resources/config/services.yml
+++ b/Resources/config/services.yml
@@ -55,7 +55,7 @@ services:
             - "@support_tools.system_info.collector.composer.lock_file"
             - "%kernel.debug%"
         # Can't tag this before v0.3 (2.5?) as it will blow up in admin UI for missing templates there
-        # And it does not look like there is anyway to add it from this package, so maybe it needs to be made extensible(?)
+        # And it does not look like there is any way to add it from this package, so maybe it needs to be made extensible(?)
         #tags: [{ name: "support_tools.system_info.collector", identifier: "ez" }]
 
     support_tools.system_info.collector.composer.lock_file:
--- a/SystemInfo/Collector/EzSystemInfoCollector.php
+++ b/SystemInfo/Collector/EzSystemInfoCollector.php
@@ -148,7 +148,7 @@ class EzSystemInfoCollector implements SystemInfoCollector
 
         // In case someone switches from TTL to BUL, make sure we only identify install as Trial if this is present,
         // as well as TTL packages
-        $hasTTLComposerRepo = in_array('https://updates.ez.no/ttl', $this->composerInfo->repositoryUrls);
+        $hasTTLComposerRepo = \in_array('https://updates.ez.no/ttl', $this->composerInfo->repositoryUrls);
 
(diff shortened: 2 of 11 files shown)


SystemInfo/Collector/JsonComposerLockSystemInfoCollector.php
```diff
@@ -101,10 +125,25 @@ public function collect()
...
         ksort($packages, SORT_FLAG_CASE | SORT_STRING);
 
-        return $this->value = new Value\ComposerSystemInfo([
-            'packages' => $packages,
-            'minimumStability' => isset($lockData['minimum-stability']) ? $lockData['minimum-stability'] : null,
-        ]);
+        return $packages;
+    }
+
+    private function extractRepositoryUrls(array $jsonData)
```
mikadamczyk: nitpick: missing docblock and return type hint

SystemInfo/Value/ComposerSystemInfo.php
```diff
@@ -32,4 +32,13 @@ class ComposerSystemInfo extends ValueObject implements SystemInfo
      * @var string|null
      */
     public $minimumStability;
+
+    /**
+     * Additional Composer repository urls used.
+     *
+     * Will contain urls used so would be possible to know if bul or ttl packages are used for instance.
+     *
+     * @var array
```
mikadamczyk: `string[]`

SystemInfo/Exception/ComposerJsonFileNotFoundException.php
```diff
@@ -0,0 +1,20 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzSupportToolsBundle\SystemInfo\Exception;
+
+use Exception;
+use eZ\Publish\Core\Base\Exceptions\NotFoundException as BaseNotFoundException;
+
+class ComposerJsonFileNotFoundException extends BaseNotFoundException
+{
+    public function __construct($path, Exception $previous = null)
```
mikadamczyk:
```suggestion
    public function __construct(string $path, Exception $previous = null)
```
if path is `string` we can add type hint

SystemInfo/Collector/EzSystemInfoCollector.php
```diff
@@ -146,15 +146,19 @@ public function collect()
...
             $ez->isEnterpise = true;
-            $ez->isTrial = $package->license === 'TTL-2.0';
+            $ez->isTrial = $hasTTLComposerRepo && $package->license === 'TTL-2.0';
             $ez->name = 'eZ Platform Enterprise';
         }
 
         if ($package = $this->getFirstPackage(self::COMMERCE_PACKAGES)) {
             $ez->isCommerce = true;
-            $ez->isTrial = $ez->isTrial || $package->license === 'TTL-2.0';
+            $ez->isTrial = $ez->isTrial || $hasTTLComposerRepo && $package->license === 'TTL-2.0';
```
glye: Aside: We have to change this when TTL version changes. Could be a `TTL-*` wildcard match, or defined as an array of valid TTLs. Disregard for now.

SystemInfo/Exception/ComposerLockFileNotFoundException.php
```diff
@@ -15,6 +15,6 @@ class ComposerLockFileNotFoundException extends BaseNotFoundException
 {
     public function __construct($path, Exception $previous = null)
     {
-        parent::__construct('Composer lockfile', $path, $previous);
+        parent::__construct('Composer.lock file', $path, $previous);
```
glye:
```suggestion
        parent::__construct('composer.lock file', $path, $previous);
```
If we're using the filename, it should be lower case.

Resources/config/services.yml
```diff
@@ -55,13 +55,14 @@ services:
...
         # Can't tag this before v0.3 (2.5?) as it will blow up in admin UI for missing templates there
-        # And it does not look like there is anway to add it from this package, so maybe it needs to be made extensible(?)
+        # And it does not look like there is anyway to add it from this package, so maybe it needs to be made extensible(?)
         #tags: [{ name: "support_tools.system_info.collector", identifier: "ez" }]
 
     support_tools.system_info.collector.composer.lock_file:
         class: "%support_tools.system_info.collector.composer.lock_file.class%"
         arguments:
             - "%kernel.root_dir%/../composer.lock"
+            - "%kernel.root_dir%/../composer.json"
```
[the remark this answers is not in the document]
andrerom: I didn't take chances of adjust the name in patch release, but indeed for a 1.1. or 2.2 it should ideally be renamed.

SystemInfo/Collector/EzSystemInfoCollector.php
andrerom: Well, this matching might fail if we start to use `property` aslo. But at least we don't risk marking non trials as trials.

PR #44, opened 2019-10-02
Remove usage of deprecated ContainerAwareCommand

emodric:


alongosz:
Tested the command with [the latest master of eZ Platform EE Demo](ezsystems/ezplatform-ee-demo@ff9e4928)

PR #52, opened 2020-04-12
Add X-Powered-By Header to promote use of our software

andrerom:
Proposal to add response listener that adds info in HTTP headers to promote that install is using eZ Platform, with param to disable it if projects wants to.

With PR I'm getting the following header locally:
```
X-Powered-By: eZ Platform Enterprise v2
```

Command-line this gives the following output:
```
$ php bin/console config:dump-reference  ezplatform_support_tools   

# Default configuration for extension with alias: "ezplatform_support_tools"
ezplatform_support_tools:

    # System info configuration, provided by "ez-support-tools" package
    system_info:

        # Control if system should generate a powered by header to promote eZ Platform usage
        powered_by:

            # Activate/Deactivate powered by header
            enabled:              true

            # Sets how much of version info is exposed in powered by. Examples:  minor: "2.5", major: "2", none: ""
            release:              major # One of "major"; "minor"; "none"



$ php bin/console debug:config  ezplatform_support_tools 

Current configuration for extension with alias "ezplatform_support_tools"
=========================================================================

ezplatform_support_tools:
    system_info:
        powered_by:
            enabled: true
            release: major
```

**Todo on merge up**:
- 3.x: Version number should rather be taken from constant available from Core package _(make sure we depend on it)_
- 3.2: Product names


alongosz:
It looks very cool, but... it's a bit security issue. We always rather aimed for a "security by obscurity" approach and not exposing software name and version to the anonymous user was a key point of that. 

AFAIK it was even raised when we made the extra effort to exclude ability to invoke front controller by passing `app.dev` directly via URI as it exposed Symfony.

We're still exposing eZ Platform in default templates though (generator meta tag), but it can be easily overridden. Hence I think opt-out, via Semantic conf, for the feature here would be great.

POV ping @glye

DependencyInjection/Compiler/SystemInfoCollectorPass.php
```diff
@@ -37,4 +44,22 @@ public function process(ContainerBuilder $container)
         $infoCollectorRegistryDef = $container->findDefinition('support_tools.system_info.collector_registry');
         $infoCollectorRegistryDef->setArguments([$infoCollectors]);
     }
+
+    private function processSystemInfo(ContainerBuilder $container)
+    {
+        if (!$container->getParameter('support_tools.promote_platform.enabled')) {
```
alongosz, line 50: Partially disregard my previous comment, I missed that.
Still, could it be exposed as a semantic config (probably mapped to this parameter later on)?
Could we keep consistent naming, starting with `ezplatform.` prefix?

andrerom: > Could we keep consistent naming, starting with ezplatform. prefix?

let's maybe wait a few months with such efforts :)


For the other remarks, I'll come back to those once the feature itself has been discussed.

glye:
Right. I second @alongosz that it should be optional (and it is). I have no opinion on _how_ it should be configured. It's on by default, this is ok, since we already expose eZ Platform otherwise, and given that it will be documented clearly, including the security implication. May need some careful diplomatic language to explain why one should / should not display this header.

In any case I think it will be difficult to completely block the fingerprinting of a reasonably complex web site as being eZ Platform, for any hacker who really wants to know. It would require the owner to more or less completely replace all default templates, among other things. If the API is exposed they can forget about it. Cloaking is mainly useful to stop robots and script kiddies, I guess, not dedicated, skillful attackers.

I consider exposing PHP version a bigger issue, because of the huge number of php web sites out there that makes it a tempting target, if it's a version with known vulnerabilities. Granted that it's controlled by php.ini and unrelated to this feature, but it would be good if the doc for the feature also mentions the php.ini setting. It would be useful if it can be set to limit the version info, like "PHP 7.3" or just "PHP 7".

andrerom:
For the whole off topic security discussion such page should then also cover `Server` and `Via`.
ez.no: `via: 1.1 varnish`
ezplatform.com: `Server: Apache/2.4.10 (Debian)`

-----

And with that I'll lock this discussion for now, until this is relevant.

DependencyInjection/Compiler/SystemInfoCollectorPass.php
andrerom: Back at this, seems @SylvainGuittard is good with doing this via optional header.

> Still, could it be exposed as a semantic config (probably mapped to this parameter later on)?

You mean plan symfony semantic config right? yeah we can do that, easier to debug.
SA config however I don't see a need for here

> Could we keep consistent naming, starting with ezplatform. prefix?

So what would you recommend as 2.5 name here for the semantic config? _(master name we can discuss on separate channels once you have moved forward on current NS / naming proposal maybe)_

andrerom:
@alongosz Got green light from @SylvainGuittard some weeks ago, so updated PR now.

```bash
$ bin/console config:dump-reference ez_support_tools

# Default configuration for extension with alias: "ez_support_tools"
ez_support_tools:

    # Control if system should generate a powered by header to promote eZ Platform usage
    powered_by:

        # Activates powered by header
        enabled:              true

        # Sets how much of version info is exposed in powered by. Examples: major: 2, minor: 2.5
        release:              minor # One of "major"; "minor"; null

        # Optional config to override the powered by naming determined by the system
        custom_name:          ''

```

DependencyInjection/Compiler/SystemInfoCollectorPass.php
```diff
@@ -37,4 +44,38 @@ public function process(ContainerBuilder $container)
         $infoCollectorRegistryDef = $container->findDefinition('support_tools.system_info.collector_registry');
         $infoCollectorRegistryDef->setArguments([$infoCollectors]);
     }
+
+    private function processSystemInfo(ContainerBuilder $container)
+    {
+        if (!$container->hasParameter('ez_support_tools.powered_by_options.enabled') ||
```
alongosz, line 50: While `ez_support_tools` is ok-ish[1] for semantic configuration - since it conforms to Symfony fallback, I wouldn't use parameter name-wise detached from eZ Platform. It gives wrong impression of separation. Support tools in current version are meant for eZ Platform only

Rather something like:
`ezplatform.support_tools.*`
or
`ezplatform.system_info.*` - while might be a bit confusing with regards to the name of the package itself, closer to its function.

---
[1] see the next diff comment

DependencyInjection/EzSystemsEzSupportToolsExtension.php
```diff
@@ -15,6 +15,16 @@
 
 class EzSystemsEzSupportToolsExtension extends Extension
 {
+    public function getAlias()
+    {
+        return 'ez_support_tools';
```
alongosz, line 20: We made arbitrary choices of extension names in the past. Looking back at this, I'm not sure if those were the right choices. The extension is meant for eZ Platform (and its variations) only ATM, so I feel like we should indicate that by
`ezplatform_support_tools` or `ezplatform_system_info`.

Not insisting though, since this would be a fallback given by Symfony if `getAlias()` was not defined (AFAIR).

DependencyInjection/EzSystemsEzSupportToolsExtension.php
```diff
@@ -15,6 +15,16 @@
...
+    }
+
+    public function getConfiguration(array $config, ContainerBuilder $container)
+    {
+        return new Configuration();
```
alongosz, line 25: Class is missing. I'm assuming this is still to be implemented. Pointing it out just in case ;)

andrerom: Not anymore :)

DependencyInjection/Compiler/SystemInfoCollectorPass.php
andrerom: Ok, yeah indeed that is what Sylvain proposed initially with:
```
ezplatform:
  system_info:
    http_header:
      expose_product_name: true
      expose_product_version: false
```

So what about:
```
# Default configuration for extension with alias: "ezplatform_system_info"
ezplatform:
  system_info:
    # Control if system should generate a powered by header to promote eZ Platform usage
    powered_by:
        # Activates powered by header
        enabled:              true
        # Sets how much of version info is exposed in powered by. Examples: major: 2, minor: 2.5
        release:              minor # One of "major"; "minor"; null
        # Optional config to override the powered by naming determined by the system
        custom_name:          ''
```

alongosz: The above is fine with me. Though maybe distinguish between `null` and empty string for `custom_name`, with the former indicating it's not an override, but not set, so eZ Platform can be set. Someone might want to override with an empty string ;)

andrerom:
Updated and rebased with fixes for review notes and some simplifications on the extension / compiler pass code.

```
(2.5)$ php bin/console config:dump-reference  ezplatform_support_tools   

# Default configuration for extension with alias: "ezplatform_support_tools"
ezplatform:

    # System info configuration, provided by "ez-support-tools" package
    system_info:

        # Control if system should generate a powered by header to promote eZ Platform usage
        powered_by:

            # Activate/Deactivate powered by header
            enabled:              true

            # Sets how much of version info is exposed in powered by. Examples: major: "2", minor: "2.5", ~: ""
            release:              minor # One of "major"; "minor"; null

            # Optional config to override the powered by naming determined by the system, set to empty string to skip name
            custom_name:          null


(2.5)$ php bin/console debug:config  ezplatform_support_tools 

Current configuration for extension with alias "ezplatform_support_tools"
=========================================================================

ezplatform_support_tools:
    system_info:
        powered_by:
            enabled: true
            release: minor
            custom_name: null
```

@alongosz Please check if this is what you had in mind here in regards to DI.

andrerom:
NOTE: As of 570f2c0 I dropped custom name, looking at this now I don't see the need for it. Either this is enabled to help promote the software it's running, or it's disabled.

@alongosz / @SylvainGuittard It's not like we had in mind using this ourselves right?

SylvainGuittard:
@andrerom I don't think we were planning to use it internally.

andrerom:
Ok, ready for re-review then 👍

glye:
Is the PR description up to date?

andrerom:
> Is the PR description up to date?

Now it is :)

DependencyInjection/EzSystemsEzSupportToolsExtension.php
```diff
@@ -23,5 +36,41 @@ public function load(array $configs, ContainerBuilder $container)
...
+                    $config['system_info']['powered_by']['release']
+                )
+            );
+        }
+    }
+
+    private function getPoweredByName(ContainerBuilder $container, ?string $release): string
+    {
+        // Autodetect product name if configured name is null (default)
+        $vendor = $container->getParameter('kernel.root_dir') . '/../vendor/';
```
alongosz, line 57: I always wonder what's gonna happen to our Product if someone sets https://getcomposer.org/doc/06-config.md#vendor-dir :)
Maybe this entire logic should happen in a Compiler pass when loaded bundles are available (AFAIK here they're not set yet)?

Resources/config/services.yml
```diff
@@ -29,6 +30,16 @@ parameters:
         support: "https://support.ez.no"
 
 services:
+    # EventSubscriber
+
+    support_tools.promote_platform.response_subscriber:
+        class: EzSystems\EzSupportToolsBundle\EventSubscriber\AddXPoweredByHeader
```
alongosz, line 36: Could you use FQCN as a service name instead?
```suggestion
    EzSystems\EzSupportToolsBundle\EventSubscriber\AddXPoweredByHeader:
```

DependencyInjection/EzSystemsEzSupportToolsExtension.php
```diff
@@ -23,5 +36,41 @@ public function load(array $configs, ContainerBuilder $container)
...
+            $name = 'eZ Platform';
+        }
+
+        // Unlike in 3.x there is no constant for version in 2.5, so while this looks hard coded it reflects composer
+        // requirements for this package version
+        If ($release === 'major') {
+            $name .= ' 2';
+        } else if ($release === 'minor') {
+            $name .= ' 2.5';
+        }
```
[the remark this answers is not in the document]
andrerom: It's stated just above ;)

> // Unlike in 3.x there is no constant for version in 2.5, so while this looks hard coded it reflects composer
> // requirements for this package version

DependencyInjection/EzSystemsEzSupportToolsExtension.php
```diff
@@ -23,5 +36,41 @@ public function load(array $configs, ContainerBuilder $container)
         $loader = new Loader\YamlFileLoader($container, new FileLocator(__DIR__ . '/../Resources/config'));
         $loader->load('services.yml');
         $loader->load('default_settings.yml');
+
+        $configuration = $this->getConfiguration($configs, $container);
+        $config = $this->processConfiguration($configuration, $configs);
+
+        if (isset($config['system_info']) && $config['system_info']['powered_by']['enabled']) {
+            $container->setParameter(
+                'ezplatform_support_tools.system_info.powered_by.name',
```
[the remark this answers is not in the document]
andrerom: Ok, but then what do we do with our `ezplatform_support_tools` alias?

ref:
```
    /**
     * Returns the recommended alias to use in XML.
     *
     * This alias is also the mandatory prefix to use when using YAML.
     *
     * This convention is to remove the "Extension" postfix from the class
     * name and then lowercase and underscore the result. So:
     *
     *     AcmeHelloExtension
     *
     * becomes
     *
     *     acme_hello
     *
     * This can be overridden in a sub-class to specify the alias manually.
     *
     * @return string The alias
     *
     * @throws BadMethodCallException When the extension name does not follow conventions
     */
    public function getAlias()
```

DependencyInjection/EzSystemsEzSupportToolsExtension.php
andrerom: Are we overthinking it here? we have plenty of code around including autoloader from vendor folder ;)

alongosz: > Are we overthinking it here? we have plenty of code around including autoloader from vendor folder ;)

:+1: True. Because indeed it would require more work, we can keep it as is for now and handle it globally when needed/possible. It's quite minor ATM.

DependencyInjection/Configuration.php
```diff
@@ -0,0 +1,57 @@
...
+ */
+final class Configuration implements ConfigurationInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getConfigTreeBuilder(): TreeBuilder
+    {
+        $treeBuilder = new TreeBuilder();
+        $rootNode = $treeBuilder->root('ezplatform');
```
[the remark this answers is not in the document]
andrerom: I§m just following what you asked for. What do you want there?

DependencyInjection/Configuration.php
```diff
@@ -0,0 +1,57 @@
...
+                            ->info('Control if system should generate a powered by header to promote eZ Platform usage')
+                            ->addDefaultsIfNotSet()
+                            ->children()
+                                ->booleanNode('enabled')
+                                    ->info('Activate/Deactivate powered by header')
+                                    ->defaultValue(true)
+                                ->end()
+                                ->enumNode('release')
+                                    ->info('Sets how much of version info is exposed in powered by. Examples: major: "2", minor: "2.5", ~: ""')
+                                    ->values(['major', 'minor', null])
```
Steveb-p, line 45: I think it should be presented in the order of amount of information that is presented, i.e. `minor` -> `major` -> none (`null`). Maybe also explicitly add `none`?

andrerom: `~` is null, but indeed it's not that obvious maybe, did you mean that we should change the line above to something like
```
>info('Sets how much of version info is exposed in powered by. Examples: major: "2", minor: "2.5", null: ""')
```

Steveb-p: Something like that. Although as I said, I'd order it by the amount of information exposed.

andrerom: ah, and you also mentioned `none`, ok so something like this then:
```
>info('Sets how much of version info is exposed in powered by. Examples:  minor: "2.5", major: "2", 'none': ""')
->values(['major', 'minor', 'none, 'null])
```

alongosz: > ah, and you also mentioned `none`, ok so something like this then:
> 
> ```
> >info('Sets how much of version info is exposed in powered by. Examples:  minor: "2.5", major: "2", none: ""')
> ->values(['major', 'minor', 'none', null])
> ```

@andrerom do we still need `null` if there's `none` option?

DependencyInjection/EzSystemsEzSupportToolsExtension.php
```diff
@@ -23,5 +36,41 @@ public function load(array $configs, ContainerBuilder $container)
...
+    {
+        // Autodetect product name if configured name is null (default)
+        $vendor = $container->getParameter('kernel.root_dir') . '/../vendor/';
+        if (is_dir($vendor . EzSystemInfoCollector::COMMERCE_PACKAGES[0])) {
+            $name = 'eZ Commerce';
+        } elseif (is_dir($vendor . EzSystemInfoCollector::ENTERPISE_PACKAGES[0])) {
+            $name = 'eZ Platform Enterprise';
+        } else {
+            $name = 'eZ Platform';
+        }
```
alongosz, line 64: So the old name still applies to 2.5 then, right? Just double checking.

alongosz (approved):
Let's stick to the config in the form:
```
ezplatform_support_tools:
    system_info:
        powered_by:
            enabled: true
            release: minor
```
for now. What I wanted needs some extra work in Kernel.

Some minor questions (but +1 regardless):

DependencyInjection/EzSystemsEzSupportToolsExtension.php
andrerom: yes, so this, and maybe others parts needs adjustments when we merge up.

For 3.2+ I was also thinking we should have Platform/Core for OSS and just DXP for all flavors of Enterprise to reflect the new naming.

DependencyInjection/Configuration.php
andrerom: No, I can drop it

DependencyInjection/EzSystemsEzSupportToolsExtension.php
andrerom: MERGE UP CHANGES:
- Product Name
- Version number from constant

DependencyInjection/Configuration.php
andrerom: Fixed in ee27032

Steveb-p (approved):

alongosz (approved):

Tests/SystemInfo/Collector/EzSystemInfoCollectorTest.php
```diff
@@ -0,0 +1,29 @@
...
+    public function testCollect(): void
+    {
+        $composerCollector = new JsonComposerLockSystemInfoCollector(
+            __DIR__ . '/_fixtures/composer.lock', __DIR__ . '/_fixtures/composer.json'
+        );
+
+        $systemInfoCollector = new EzSystemInfoCollector($composerCollector);
+        $systemInfo = $systemInfoCollector->collect();
+        self::assertSame(EzSystemInfo::PRODUCT_NAME_OSS, $systemInfo->name);
+        self::assertSame("2.5", $systemInfo->release);
```
alongosz, line 27: CS:
```suggestion
        self::assertSame('2.5', $systemInfo->release);
```

alongosz (changes requested):

Tests/SystemInfo/Collector/EzSystemInfoCollectorTest.php
alongosz: @andrerom Travis is actually failing on that ^

alongosz (approved):

mnocon (commented):
Tested on all metarepositories with v2( ezplatform, ezplatform-ee, ezcommerce), looks good!

One thing that I haven't seen mentioned here: IMHO we should not only mention in the doc (on a special page), but also in the update section - so that people updating from previous v2 versions can make an informed decision about keeping it enabled.

No approve yet, because fixer is failing: https://travis-ci.org/github/ezsystems/ez-support-tools/jobs/738956580

and I can see it used to pass, even for earlier commits in this PR.

mnocon:
About the fixer thing: it's caused by Composer 2 (https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/5238)

AFAICS the options are:
1) update to php-cs-fixer 2.16.7
2) keep using Composer1 for CS job
Optional: switch to `ezsystems/ezplatform-code-style`?

andrerom:
> About the fixer thing: it's caused by Composer 2

Would align with master here, but maybe it's intended to switch to use `ezsystems/ezplatform-code-style`, if so, any example repo

mnocon:
Yes, we intend to switch to ezplatform-code-style eventually to make maintenence easier.
Example PR switching: https://github.com/ezsystems/ezplatform-page-builder/pull/424/files (use 0.1.0 for the old file header, 0.2.0 uses `Ibexa` file header).

andrerom:
Switched to `ezplatform-code-style` in 1.0, already the case for 2.x so din't carry the change there as it as different config for some reason. It took roughly 10min so we don't have to waste time on workarounds like hardcoding composer v1 ;)

mnocon:
> Unsure why Travis don't pick up the change though.

Travis had issues over the weekend and didn't start builds when it should have - I don't think we can restart it in from the GUI, can you push your changes again to trigger it again?

andrerom:
@mnocon Added small change here to only show major version, so that should implicit trigger new test run as well.

micszo (approved):
Verified across different meta repositories with 2.5.
Checked with symfony http cache and Varnish.

LGTM.

One curious thing: on Commerce the header is displayed in one row (together with PHP, comma-separated), on OSS and EE it's it's displayed in a separate row. 🤷‍♂️

andrerom:
> One curious thing: on Commerce the header is displayed in one row (together with PHP, comma-separated), on OSS and EE it's displayed in a separate row. 🤷‍♂️


I would be-live this is done by the web server, unsure how we can control it.

PR #53, opened 2020-07-03
EZP-31719: Incorrect end of life information on Dashboard

glye:
https://jira.ez.no/browse/EZP-31719

Updated release, EOM and EOL dates based on 3.0 tag date and https://support.ez.no/Public/Service-Life

The 3.0 life cycle has been squashed here a lot compared to how it was. Suggestions? If we agree to extend it, I will update https://support.ez.no/Public/Service-Life too.

Need to fix the HTML showing up in the alert box too, but that's in admin-ui repo.
PR: https://github.com/ezsystems/ezplatform-admin-ui/pull/1421

### DIFF
--- a/SystemInfo/Collector/EzSystemInfoCollector.php
+++ b/SystemInfo/Collector/EzSystemInfoCollector.php
@@ -37,8 +37,8 @@ class EzSystemInfoCollector implements SystemInfoCollector
         '2.3' => '2018-09-20T23:59:59+00:00',
         '2.4' => '2018-12-20T23:59:59+00:00',
         '2.5' => '2019-03-29T16:59:59+00:00',
-        '3.0' => '2019-12-20T23:59:59+00:00', // Estimate at time of writing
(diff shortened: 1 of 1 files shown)


lserwatka:
Could you merge it up?

glye:
@lserwatka Merged. Will you release it?

lserwatka:
Yes, will do.

micszo:
Q: when the EOM date has passed (and the EOL date hasn't) shouldn't the message be

> Unfortunately x.y.z open source version has reached end of maintenance

instead of

> Unfortunately x.y.z open source version has reached end of life

?

I've been playing with the dates in this file and it seems in both cases it stays `end of life`.
Could someone check the logic in the code?

andrerom:
@micszo AFAIK it's on purpose, as strictly speaking community does not receive any fixes after EOM as that is purely  for enterprise customers.

PR #56, opened 2020-07-29
Update v3.1 release/lifetime dates

glye:
v3.1 dates can be nailed now.

v3.2 release is expected at September 28th, but EOM/EOL dates are not known. Afaik it's not clear if 3.2 will be FT or LTS, so for now I haven't set these. Let me know if you have more info.


andrerom:
This _can_ be applied to 2.1 branch instead, 1.0 does not support use with 3.x

glye:
Seems consistent with 2.0 and 2.1 to just leave the estimates as is.
2.1 pr in https://github.com/ezsystems/ez-support-tools/pull/57

PR #73, opened 2020-11-04
Update/correct 3.1/3.2/3.3 dates

glye:
Update 3.1/3.2/3.3 release dates in v3.2+

- Correct 3.1 dates to what's already set in that branch
- Update 3.2 dates based on 4 months maintenance and 2 months security
- Add 3.3 dates, also 4 + 2

### DIFF
--- a/src/bundle/SystemInfo/Collector/IbexaSystemInfoCollector.php
+++ b/src/bundle/SystemInfo/Collector/IbexaSystemInfoCollector.php
@@ -34,7 +34,8 @@ class IbexaSystemInfoCollector implements SystemInfoCollector
         '2.5' => '2019-03-29T16:59:59+00:00',
         '3.0' => '2020-04-02T23:59:59+00:00',
         '3.1' => '2020-07-15T23:59:59+00:00',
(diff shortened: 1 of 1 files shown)


src/bundle/SystemInfo/Collector/IbexaSystemInfoCollector.php
```diff
@@ -34,7 +34,8 @@ class IbexaSystemInfoCollector implements SystemInfoCollector
         '2.5' => '2019-03-29T16:59:59+00:00',
         '3.0' => '2020-04-02T23:59:59+00:00',
         '3.1' => '2020-07-15T23:59:59+00:00',
-        '3.2' => '2020-09-20T23:59:59+00:00', // Estimate at time of writing
+        '3.2' => '2020-10-23T23:59:59+00:00',
+        '3.3' => '2020-12-30T23:59:59+00:00', // Estimate at time of writing
```
andrerom, line 38: It will be before this (due to xmas), check for planned date here maybe

glye: I was being pessimistic on purpose here, but LS confirms, so 🤷‍♂️

PR #74, opened 2020-11-15
[DX] Setup short forward compatible 'ibexa:info' aliases for system info dump command

andrerom:
Forward compatibility, as we often instructs customer to run this, and don't always know which version they are on (upgrades and so on.)

This also simplify instructions for how to dump info about the system.

### DIFF
--- a/Command/SystemInfoDumpCommand.php
+++ b/Command/SystemInfoDumpCommand.php
@@ -48,7 +48,6 @@ class SystemInfoDumpCommand extends ContainerAwareCommand
         $this
             ->setName('ez-support-tools:dump-info')
             ->setAliases([
-                'ez:info',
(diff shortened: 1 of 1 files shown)


konradoboza (approved):

glye (approved):
Reposting updated slack comment:

| Not sure I like creating new stuff using "ez" now. "ibexa:info" is ok I think. Granted, if "ibexa:" commands don't exist in kernel yet, then "ez:" is surely fine. Also fine for 3.1 and older.

So if "ez:" commands are still maintained in 3.2, then it's fine for us as well. We don't have to be holier than ~thou~ engineering.

mateuszdebinski (approved):

mateuszbieniek (approved):

PR #76, opened 2020-12-17
EZP-32243: Exposed repository object counts in database tab

konradoboza:
JIRA: https://issues.ibexa.co/browse/EZP-32243

Excerpt from the ticket:
>The database tab should also contain information about current repository metrics like:
>- published content objects count,
>- users count,
>- drafts count,
>- content types count.
>
>To be expanded further in the next steps, further ideas can be found in https://github.com/ezsystems/support-site/issues/108.

**Note:** the failing job is related to not updated Behat tests, I will do a followup once this PR gets accepted.

Zrzut ekranu 2020-12-22 o 14 50 09

Related PR:
https://github.com/ezsystems/ezplatform-admin-ui/pull/1681

- [x] adapt existing unit tests,
- [x] followup in AdminUI for Behat scenarios.


src/lib/Storage/DoctrineGateway.php
```diff
@@ -0,0 +1,102 @@
...
+    /** @var \Doctrine\DBAL\Platforms\AbstractPlatform */
+    private $databasePlatform;
+
+    public function __construct(Connection $connection)
+    {
+        $this->connection = $connection;
+        $this->databasePlatform = $connection->getDatabasePlatform();
+    }
+
+    public function getContentObjectsCount(bool $includeDrafts = false): int
```
andrerom, line 36: Could change logic to `bool $onlyPublished = true`, as there is also archived state (trash) right?

src/lib/Storage/DoctrineGateway.php
```diff
@@ -0,0 +1,102 @@
...
+    {
+        $queryBuilder = $this->connection->createQueryBuilder();
+        $queryBuilder
+            ->select($this->databasePlatform->getCountExpression('contentobject_id'))
+            ->from(self::USER_TABLE);
+
+        return (int) $queryBuilder->execute()->fetchColumn();
+    }
+
+    public function getDraftsCount(): int
```
andrerom, line 62: Hmm, good to have drafts, but might be even more interesting to know how many versions there are in the system in total as it is what will affects performance of SQL queries more.

konradoboza: Added in https://github.com/ezsystems/ez-support-tools/pull/76/commits/2709ba354dd095baa067987521e3c67bef7e60e4.

src/lib/Storage/DoctrineGateway.php
konradoboza: I added a condition ruling out trashed items. As we present this statistic as "Published content objects count" I would prefer staying with `bool includeDrafts = false`. Changed in https://github.com/ezsystems/ez-support-tools/pull/76/commits/2709ba354dd095baa067987521e3c67bef7e60e4.

andrerom:
@mnocon @micszo Should maybe the Behat spec features be moved into this package instead?

andrerom (approved):

mnocon:
@andrerom yes, I plan to do that in January once 3.3 is released (I have a refactor started that allows us to decouple tests easily).

src/bundle/Resources/views/themes/admin/system_info/repository.html.twig
```diff
@@ -0,0 +1,60 @@
+{% trans_default_domain "systeminfo" %}
+
+<!-- Tab name: {{ 'tab.name.repository'|trans|desc('Repository') }} -->
+
+<section class="ez-fieldgroup">
+    <h2 class="ez-fieldgroup__name">{{ 'repository'|trans|desc('Repository') }}</h2>
+    <div class="ez-fieldgroup__content">
+        <table class="table ez-table--list">
+            <tbody>
+            <tr class="ez-table__row">
```
dew326, line 10: should be indented one more

src/bundle/Resources/views/themes/admin/system_info/repository.html.twig
```diff
@@ -0,0 +1,60 @@
...
+            </tbody>
+        </table>
+    </div>
+</section>
+<section class="ez-fieldgroup">
+    <h2 class="ez-fieldgroup__name">{{ 'database'|trans|desc('Database') }}</h2>
+    <div class="ez-fieldgroup__content">
+        <table class="table ez-table--list">
+            <tbody>
+            <tr class="ez-table__row">
```
dew326, line 41: should be indented one more

dew326 (approved):

src/bundle/Resources/config/services.yaml
```diff
@@ -10,38 +10,37 @@ parameters:
...
-    support_tools.system_info.collector.database.doctrine.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\DoctrineDatabaseSystemInfoCollector
+    support_tools.system_info.collector.repository.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\RepositorySystemInfoCollector
     support_tools.system_info.collector.system.ibexa.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\IbexaSystemInfoCollector
     support_tools.system_info.collector.hardware.ezc.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\EzcHardwareSystemInfoCollector
     support_tools.system_info.collector.php.ezc.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\EzcPhpSystemInfoCollector
     support_tools.system_info.collector.symfony.kernel.config.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\ConfigurationSymfonyKernelSystemInfoCollector
     support_tools.system_info.output_format.json.class: EzSystems\EzSupportToolsBundle\SystemInfo\OutputFormat\JsonOutputFormat
+    support_tools.system_info.gateway.doctrine.class: EzSystems\EzSupportTools\Storage\DoctrineGateway
```
alongosz, line 19: the same comment as above

src/bundle/SystemInfo/Value/RepositorySystemInfo.php
```diff
@@ -48,4 +48,11 @@ class DatabaseSystemInfo extends ValueObject implements SystemInfo
      * @var string
      */
     public $username;
+
+    /**
+     * RepositoryStatistics contains counts of content objects, users etc.
+     *
+     * @var RepositoryStatistics
```
alongosz, line 55: FQCN please

src/bundle/Tests/SystemInfo/Collector/RepositorySystemInfoCollectorTest.php
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * @var RepositorySystemInfoCollector
+     */
+    private $repositoryCollector;
+
+    protected function setUp(): void
+    {
+        $this->dbalConnectionMock = $this->getMockBuilder('Doctrine\DBAL\Connection')->disableOriginalConstructor()->getMock();
+        $this->dbalPlatformMock = $this->getMockBuilder('Doctrine\DBAL\Platforms\MySqlPlatform')->getMock();
```
alongosz, line 40: Use `::class` instead of strings

src/lib/Storage/DoctrineGateway.php
```diff
@@ -0,0 +1,114 @@
...
+    /** @var \Doctrine\DBAL\Connection */
+    private $connection;
+
+    /** @var \Doctrine\DBAL\Platforms\AbstractPlatform */
+    private $databasePlatform;
+
+    public function __construct(Connection $connection)
+    {
+        $this->connection = $connection;
+        $this->databasePlatform = $connection->getDatabasePlatform();
```
alongosz, line 32: Needs to be fetched lazily because when container is compiled, db might not be available yet.

alongosz (changes requested):
Partial review:

src/lib/Storage/DoctrineGateway.php
konradoboza: Not sure about this: if we still aim to use `final` scope for metrics, they cannot be marked as `lazy`.

konradoboza:
I reworked the current approach by introducing tagged metrics instead of one, generic gateway. Side: the failing job still refers to Behat tests, ref: https://travis-ci.com/github/ezsystems/ez-support-tools/jobs/463901364.

andrerom:
Review ping @alongosz

src/bundle/Resources/config/services.yaml
```diff
@@ -10,38 +10,35 @@ parameters:
...
-        doc: "https://doc.ezplatform.com"
-        update: "https://doc.ezplatform.com/en/latest/releases/updating_ez_platform/"
+        consulting_service: "https://www.ibexa.co/services/training"
+        ee_product: "https://www.ibexa.co/products"
+        install_ee: "https://doc.ibexa.co/en/{ez.release}/getting_started/install_ez_enterprise/"
+        doc: "https://doc.ibexa.co"
+        update: "https://doc.ibexa.co/en/latest/updating/updating_ez_platform/"
         gpl_faq: "https://www.gnu.org/licenses/old-licenses/gpl-2.0-faq.en.html#GPLModuleLicense"
-        support: "https://support.ez.no"
+        support: "https://support.ibexa.co"
```
andrerom, line 32: @alongosz I believe you'r change on the other PR will conflict with this. As this is overlapping, maybe you should take over this given Konrad is off until 11th.

alongosz: > I believe you'r change on the other PR will conflict with this. As this is overlapping, maybe you should take over this given Konrad is off until 11th.

Good catch. Moved to a separate commit to be merged as a part of #80 and easily rebase here.

src/lib/Storage/DoctrineGateway.php
alongosz: > Not sure about this: if we still aim to use `final` scope for metrics, they cannot be marked as `lazy`.

@konradoboza  What I've meant is that `getDatabasePlatform` cannot be called in the constructor. It's to heavy and when db is not configured yet (composer create-project) it's likely to crash (depends on how the container is being built atm)

Fixed by common base, since you needed to pass `ezpublish.spi.persistence.connection` to make it repository-aware.
ref.: `\EzSystems\EzSupportTools\Storage\Metrics\RepositoryConnectionAwareMetrics`

src/bundle/Resources/config/services.yaml
```diff
@@ -109,8 +106,45 @@ services:
...
         tags:
             - { name: "support_tools.system_info.output_format", format: "json" }
+
+    # Gateways
+    EzSystems\EzSupportTools\Storage\MetricsProvider:
+        alias: EzSystems\EzSupportTools\Storage\AggregateMetricsProvider
+
+    EzSystems\EzSupportTools\Storage\AggregateMetricsProvider:
+        arguments:
+            $metrics: !tagged_iterator { tag: ibexa.system_info.metrics, index_by: identifier }
```
Steveb-p, line 120: Just a note that `!tagged_locator` / `ServiceLocator` can be used here instead, since we're looking for a particular service anyway each time, based on identifier.

src/bundle/Resources/config/services.yaml
```diff
@@ -109,8 +106,45 @@ services:
...
+
+    EzSystems\EzSupportTools\Storage\Metrics\RepositoryConnectionAwareMetrics:
+        abstract: true
+        arguments:
+            $connection: '@ezpublish.persistence.connection'
+
+    EzSystems\EzSupportTools\Storage\Metrics\PublishedContentObjectsCountMetrics:
+        autowire: true
+        tags:
+            - { name: ibexa.system_info.metrics, identifier: published }
```
Steveb-p, line 130: We can also consider using `!php/const`, so the identifier usage can be more easily found throughout codebase if we ever change them.

https://symfony.com/blog/new-in-symfony-3-2-php-constants-in-yaml-files

src/bundle/Resources/config/services.yaml
alongosz: > Just a note that `!tagged_locator` / `ServiceLocator` can be used here instead, since we're looking for a particular service anyway each time, based on identifier.

So something like that: 733f3b8f? If so then maybe you've just found one useful application of locators in favor of iterators. I don't see even static analysis complain on return type, which is good.

src/bundle/Resources/config/services.yaml
alongosz: Fixed via e9785ac0 though not sold on readability of this solution. I hope further IDE upgrades would provide at least refactoring for that.

src/bundle/Resources/config/services.yaml
Steveb-p: Pretty much yeah.

This has the added benefit of **not** initializing services in this locator until they are actually used.

src/lib/Storage/AggregateMetricsProvider.php
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * @throws \EzSystems\EzSupportToolsBundle\SystemInfo\Exception\MetricsNotFoundException
+     */
+    public function provideMetrics(string $identifier): Metrics
+    {
+        try {
+            return $this->metricsLocator->get($identifier);
+        } catch (ServiceNotFoundException $e) {
+            throw new MetricsNotFoundException($identifier);
```
Steveb-p, line 36:
```suggestion
            throw new MetricsNotFoundException($identifier, $e);
```

alongosz: Good catch, committed.

Steveb-p (approved):
LGTM, although as discussed with Andrew it requires further changes to be properly user-extendable.

tomaszszopinski (approved):
QA approved on Ibexa DXP (flex) & Ibexa DXP oss with diff.

alongosz (approved):

PR #86, opened 2021-01-22
EZP-32307: Fixed check version stability

ciastektk:
https://issues.ibexa.co/browse/EZP-32307

This PR provides implemented `VersionStabilityChecker` for displaying valid version stability. Since `Ibexa/website-skeleton` (https://github.com/ibexa/website-skeleton/blob/main/composer.json#L6) provides `minimum-stability` set to `dev` all versions are marked as ‘`dev`’. Now version stability is based on `ibexa/oss` version because this package is installed for all Ibexa Platform versions. For BC break reason old check based on composer.lock minimum stability is set as callback in case when Ibexa/oss is not found.

### DIFF
--- a/src/bundle/SystemInfo/Collector/JsonComposerLockSystemInfoCollector.php
+++ b/src/bundle/SystemInfo/Collector/JsonComposerLockSystemInfoCollector.php
@@ -6,6 +6,9 @@
  */
 namespace EzSystems\EzSupportToolsBundle\SystemInfo\Collector;
 
+use Composer\InstalledVersions;
+use EzSystems\EzSupportTools\Value\Stability;
+use EzSystems\EzSupportTools\VersionStability\VersionStabilityChecker;
 use EzSystems\EzSupportToolsBundle\SystemInfo\Exception;
 use EzSystems\EzSupportToolsBundle\SystemInfo\Value\ComposerSystemInfo;
 use EzSystems\EzSupportToolsBundle\SystemInfo\Value\ComposerPackage;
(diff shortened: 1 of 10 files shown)


src/bundle/SystemInfo/Collector/JsonComposerLockSystemInfoCollector.php
```diff
@@ -171,4 +174,25 @@ private static function setNormalizedVersion(ComposerPackage $package): void
...
+                return $this->getIbexaOssStability($package['version']);
+            }
+        }
+
+        return $this->getMinimumStability($lockData);
+    }
+
+    private function getIbexaOssStability(string $version): string
+    {
+        return $version === self::DEV_VERSION ? self::STABILITIES[20] : self::STABILITIES[0];
```
alongosz, line 191: This is neither stable nor complete. As you can see `self::STABILITIES` contains multiple stability constraints. For RC packages we need to display "RC", for Beta - "Beta" and so on.
Moreover while right `$version === 'dev-master'` is correct for the latest release, for earlier ones it will become e.g. `3.2.x-dev` or `3.4.x-dev`. To retrieve that information we should rather use PHP API from `composer/composer` package. I wonder if it contains some more abstract way to get a stability of a package - TBD.

alongosz (changes requested):

src/bundle/Tests/SystemInfo/Collector/JsonComposerLockSystemInfoCollectorTest.php
```diff
@@ -82,7 +95,11 @@ public function testCollectLockFileNotFound()
     {
         $this->expectException(ComposerLockFileNotFoundException::class);
 
-        $composerCollectorNotFound = new JsonComposerLockSystemInfoCollector(__DIR__ . '/_fixtures/snafu.lock', __DIR__ . '/_fixtures/composer.json');
+        $composerCollectorNotFound = new JsonComposerLockSystemInfoCollector(
+            $this->composerInstalledVersions,
+            __DIR__ . '/_fixtures/snafu.lock',
```
alongosz, line 100: :joy:

alongosz (approved):
Good responsibility separation!

tomaszszopinski (approved):
QA approved on Ibexa oss/content/ezp/commerce stable 3.3@rc with patch.

PR #89, opened 2021-02-17
EZP-32348: Exposed services used within the project

konradoboza:
JIRA ticket: https://issues.ibexa.co/browse/EZP-32348

> The idea is to add another tab containing information about currently used services:
> - search engine,
> - HTTP cache reverse proxy,
> - persistence cache adapter.
>
> To be expanded further in the next steps, more ideas can be found in https://github.com/ezsystems/support-site/issues/108.

Zrzut ekranu 2021-02-17 o 13 15 27

### DIFF
--- a/src/bundle/Resources/config/services.yaml
+++ b/src/bundle/Resources/config/services.yaml
@@ -14,7 +14,6 @@ parameters:
     support_tools.system_info.collector.hardware.ezc.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\EzcHardwareSystemInfoCollector
     support_tools.system_info.collector.php.ezc.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\EzcPhpSystemInfoCollector
     support_tools.system_info.collector.symfony.kernel.config.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\ConfigurationSymfonyKernelSystemInfoCollector
-    support_tools.system_info.collector.services.config.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\ServicesSystemInfoCollector
     support_tools.system_info.output_format.json.class: EzSystems\EzSupportToolsBundle\SystemInfo\OutputFormat\JsonOutputFormat
     ezplatform_support_tools.system_info.powered_by.name: ''
     ibexa.system_info.ibexa_url_list:
@@ -113,8 +112,7 @@ services:
         tags:
             - { name: "support_tools.system_info.collector", identifier: "symfony_kernel" }
 
-    support_tools.system_info.collector.services.config:
-        class: "%support_tools.system_info.collector.services.config.class%"
+    EzSystems\EzSupportToolsBundle\SystemInfo\Collector\ServicesSystemInfoCollector:
         arguments:
             - '@parameter_bag'
         tags:
--- a/src/bundle/Resources/config/services.yaml
+++ b/src/bundle/Resources/config/services.yaml
@@ -14,6 +14,7 @@ parameters:
     support_tools.system_info.collector.hardware.ezc.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\EzcHardwareSystemInfoCollector
     support_tools.system_info.collector.php.ezc.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\EzcPhpSystemInfoCollector
     support_tools.system_info.collector.symfony.kernel.config.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\ConfigurationSymfonyKernelSystemInfoCollector
+    support_tools.system_info.collector.services.config.class: EzSystems\EzSupportToolsBundle\SystemInfo\Collector\ServicesSystemInfoCollector
     support_tools.system_info.output_format.json.class: EzSystems\EzSupportToolsBundle\SystemInfo\OutputFormat\JsonOutputFormat
     ezplatform_support_tools.system_info.powered_by.name: ''
     ibexa.system_info.ibexa_url_list:
@@ -111,6 +112,13 @@ services:
             - "%kernel.bundles%"
         tags:
             - { name: "support_tools.system_info.collector", identifier: "symfony_kernel" }
+
+    support_tools.system_info.collector.services.config:
+        class: "%support_tools.system_info.collector.services.config.class%"
+        arguments:
+            - '@parameter_bag'
+        tags:
+            - { name: "support_tools.system_info.collector", identifier: "services" }
 
     # SystemInfoOutputFormats
     support_tools.system_info.output_format.json:
--- a/src/bundle/SystemInfo/Collector/ServicesSystemInfoCollector.php
+++ b/src/bundle/SystemInfo/Collector/ServicesSystemInfoCollector.php
@@ -28,8 +28,6 @@ class ServicesSystemInfoCollector implements SystemInfoCollector
 
     /**
      * Collects information about the utilized services.
-     *
-     * @return \EzSystems\EzSupportToolsBundle\SystemInfo\Value\ServicesSystemInfo
      */
     public function collect(): ServicesSystemInfo
     {
--- a/src/bundle/SystemInfo/Collector/ServicesSystemInfoCollector.php
+++ b/src/bundle/SystemInfo/Collector/ServicesSystemInfoCollector.php
@@ -1,0 +1,42 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzSupportToolsBundle\SystemInfo\Collector;
+
+use EzSystems\EzSupportToolsBundle\SystemInfo\Value\ServicesSystemInfo;
+use Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface;
+
+/**
+ * Collects information about the services used within the project.
+ */
+class ServicesSystemInfoCollector implements SystemInfoCollector
+{
+    private const SEARCH_ENGINE_CONFIG_KEY = 'search_engine';
+    private const HTTP_CACHE_CONFIG_KEY = 'purge_type';
+    private const PERSISTENCE_CACHE_CONFIG_KEY = 'cache_pool';
+
+    /** @var \Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface */
+    private $parameterBag;
+
+    public function __construct(ParameterBagInterface $parameterBag)
+    {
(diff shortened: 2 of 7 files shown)


src/bundle/Resources/config/services.yaml
```diff
@@ -112,6 +113,13 @@ services:
         tags:
             - { name: "support_tools.system_info.collector", identifier: "symfony_kernel" }
 
+    support_tools.system_info.collector.services.config:
```
ViniTou, line 116: Can we just use class name as service name and drop `.class` param?

src/bundle/Tests/SystemInfo/Collector/ServicesSystemInfoCollectorTest.php
```diff
@@ -0,0 +1,63 @@
...
+
+class ServicesSystemInfoCollectorTest extends TestCase
+{
+    /**
+     * @var \Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface|\PHPUnit\Framework\MockObject\MockObject
+     */
+    private $parameterBagMock;
+
+    /**
+     * @var ServicesSystemInfoCollector
```
ViniTou, line 22:
```suggestion
     * @var \EzSystems\EzSupportToolsBundle\SystemInfo\Collector\ServicesSystemInfoCollector
```

src/bundle/SystemInfo/Collector/ServicesSystemInfoCollector.php
```diff
@@ -0,0 +1,42 @@
...
+
+    public function __construct(ParameterBagInterface $parameterBag)
+    {
+        $this->parameterBag = $parameterBag;
+    }
+
+    /**
+     * Collects information about the utilized services.
+     *
+     * @return \EzSystems\EzSupportToolsBundle\SystemInfo\Value\ServicesSystemInfo
```
adamwojs, line 32: Nitpick: `@return` tag is redundant here

src/bundle/Tests/SystemInfo/Collector/ServicesSystemInfoCollectorTest.php
```diff
@@ -0,0 +1,63 @@
...
+    public function testCollect()
+    {
+        $expected = new ServicesSystemInfo([
+            'searchEngine' => 'solr',
+            'httpCacheProxy' => 'varnish',
+            'persistenceCacheAdapter' => 'cache.adapter.memcached',
+        ]);
+
+        $this->parameterBagMock
+            ->expects($this->at(0))
```
Steveb-p, line 44: Using `Assert::at` is deprecated. Use `willReturnConsecutive` instead.

You can also make assertions about what is passed as argument to `ParameterBag` instance with `withConsecutive`.

src/bundle/SystemInfo/Value/ServicesSystemInfo.php
```diff
@@ -0,0 +1,42 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzSupportToolsBundle\SystemInfo\Value;
+
+use eZ\Publish\API\Repository\Values\ValueObject;
+
+/**
+ * Value for information about services used within the project.
+ */
+class ServicesSystemInfo extends ValueObject implements SystemInfo
```
Steveb-p, line 14:
```suggestion
final class ServicesSystemInfo extends ValueObject implements SystemInfo
```

src/bundle/SystemInfo/Collector/ServicesSystemInfoCollector.php
```diff
@@ -0,0 +1,40 @@
...
+ */
+namespace EzSystems\EzSupportToolsBundle\SystemInfo\Collector;
+
+use EzSystems\EzSupportToolsBundle\SystemInfo\Value\ServicesSystemInfo;
+use Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface;
+
+/**
+ * Collects information about the services used within the project.
+ */
+class ServicesSystemInfoCollector implements SystemInfoCollector
```
Steveb-p, line 15:
```suggestion
final class ServicesSystemInfoCollector implements SystemInfoCollector
```

src/bundle/Tests/SystemInfo/Collector/ServicesSystemInfoCollectorTest.php
```diff
@@ -0,0 +1,63 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\EzSupportToolsBundle\Tests\SystemInfo\Collector;
+
+use EzSystems\EzSupportToolsBundle\SystemInfo\Collector\ServicesSystemInfoCollector;
+use EzSystems\EzSupportToolsBundle\SystemInfo\Value\ServicesSystemInfo;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\DependencyInjection\ParameterBag\ParameterBagInterface;
+
+class ServicesSystemInfoCollectorTest extends TestCase
```
Steveb-p, line 14:
```suggestion
final class ServicesSystemInfoCollectorTest extends TestCase
```

Steveb-p (changes requested):

src/bundle/Tests/SystemInfo/Collector/ServicesSystemInfoCollectorTest.php
```diff
@@ -0,0 +1,63 @@
...
+    protected function setUp(): void
+    {
+        $this->parameterBagMock = $this->createMock(ParameterBagInterface::class);
+        $this->servicesCollector = new ServicesSystemInfoCollector($this->parameterBagMock);
+    }
+
+    /**
+     * @covers \EzSystems\EzSupportToolsBundle\SystemInfo\Collector\RepositorySystemInfoCollector::collect()
+     */
+    public function testCollect()
```
adamwojs, line 35:
```suggestion
    public function testCollect(): void
```

src/bundle/Tests/SystemInfo/Collector/ServicesSystemInfoCollectorTest.php
```diff
@@ -0,0 +1,63 @@
...
+        $expected = new ServicesSystemInfo([
+            'searchEngine' => 'solr',
+            'httpCacheProxy' => 'varnish',
+            'persistenceCacheAdapter' => 'cache.adapter.memcached',
+        ]);
+
+        $this->parameterBagMock
+            ->expects($this->at(0))
+            ->method('get')
+            ->willReturn($expected->searchEngine);
```
adamwojs, line 46: You are missing here `->with(...)` otherwise this test doesn't make sense.

src/bundle/SystemInfo/Collector/ServicesSystemInfoCollector.php
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * Collects information about the utilized services.
+     */
+    public function collect(): ServicesSystemInfo
+    {
+        return new ServicesSystemInfo([
+            'searchEngine' => $this->parameterBag->get(self::SEARCH_ENGINE_CONFIG_KEY),
+            'httpCacheProxy' => $this->parameterBag->get(self::HTTP_CACHE_CONFIG_KEY),
+            'persistenceCacheAdapter' => $this->parameterBag->get(self::PERSISTENCE_CACHE_CONFIG_KEY),
```
alongosz, line 37: Search engine and persistence cache would show incorrect information in case the given admin interface used non-default repository in multi-repo scenario. Search engine is configurable on a Repository and a Repository is configurable for SiteAccess. Persistence cache adapter is configurable per SiteAccess via `cache_service_name`.

alongosz (changes requested):

src/bundle/Tests/SystemInfo/Collector/ServicesSystemInfoCollectorTest.php
```diff
@@ -0,0 +1,63 @@
...
+            ->willReturn($expected->httpCacheProxy);
+
+        $this->parameterBagMock
+            ->expects($this->at(2))
+            ->method('get')
+            ->willReturn($expected->persistenceCacheAdapter);
+
+        $value = $this->servicesCollector->collect();
+
+        self::assertInstanceOf(ServicesSystemInfo::class, $value);
```
adamwojs, line 60: This assertion is redundant: correct type is guaranteed on method definition level.

src/bundle/Tests/SystemInfo/Collector/ServicesSystemInfoCollectorTest.php
```diff
@@ -0,0 +1,63 @@
...
+     */
+    public function testCollect()
+    {
+        $expected = new ServicesSystemInfo([
+            'searchEngine' => 'solr',
+            'httpCacheProxy' => 'varnish',
+            'persistenceCacheAdapter' => 'cache.adapter.memcached',
+        ]);
+
+        $this->parameterBagMock
```
mikadamczyk, line 43: We should not check or rely on the order of elements here. The returned value probably should not be a part of what we expect in an assertion. `willReturnMap` or `willReturnCallback` could be useful

src/bundle/SystemInfo/Collector/ServicesSystemInfoCollector.php
alongosz, line 37: Architecture remark: this violates open-closed principle because if you'd need to add something new you'd need to modify this code (and additionaly in a repetitive manner). Some sort subscriber (not necessarily in SF terms) or visitor should be applied here with a strategy per the given entry so adding new item does not involve changing that code.

alongosz (changes requested):

webhdx (changes requested):
Indeed a welcomed feature but sadly this is not going to work in all cases.

You are relying on Container parameters to check what services are being used. The parameters are not what really make particular change in the setup - these are only used as a helper to pass values to the configuration i.e.:
```yaml

ezplatform:
    # HttpCache settings, By default 'local' (Symfony HttpCache Proxy), by setting it to 'varnish' you can point it to Varnish
    http_cache:
        purge_type: '%purge_type%'

    # Repositories configuration, setup default repository to support solr if enabled
    repositories:
        default:
            storage: ~
            search:
                engine: '%search_engine%'
                connection: default
```

So nothing stops you from changing search engine directly in the configuration which means `search_engine` parameter no longer matters and your collector will collect invalid data.

src/bundle/Tests/SystemInfo/Collector/ServicesSystemInfoCollectorTest.php
Steveb-p: This would be detected by phpstan with phpunit-extension installed :)

// advert mode off ;)

konradoboza:
This improvement will be reworked for https://github.com/ibexa/support-tools repository, closing.

PR #98, opened 2021-06-22
IBX-497: Redesign tabs for 4.0

tischsoic:
JIRA: https://issues.ibexa.co/browse/IBX-497

main PR: https://github.com/ezsystems/ezplatform-admin-ui/pull/1780

### DIFF
--- a/src/bundle/Resources/views/themes/admin/ui/tab/system_info.html.twig
+++ b/src/bundle/Resources/views/themes/admin/ui/tab/system_info.html.twig
@@ -1,12 +1,9 @@
-{% set tabs_to_show = [] %}
-{% for tab in tabs %}
-    {% set tabs_to_show = tabs_to_show|merge([{ 
-        id: group ~ '-' ~ tab.identifier,
-        label: tab.name,
-        content: tab.view,
-        active: loop.first,
-    }]) %}
-{% endfor %}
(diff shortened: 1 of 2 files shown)


src/bundle/Resources/views/themes/admin/ui/tab/system_info.html.twig
```diff
@@ -1,26 +1,13 @@
...
-        {% set active = loop.first %}
-
-        <div class="tab-pane{% if active %} active{% endif %}" id="ez-tab-{{ id }}" role="tabpanel"
-             aria-labelledby="ez-tab-label-{{ group }}">
-            {{ tab.view|raw }}
-        </div>
-    {% endfor %}
-</div>
+{% set tabs_to_show = [] %}
+{% for tab in tabs %}
```
GrabowskiM, line 2: I think it would be cleaner to use map here
```
set tabs_to_show = tabs|map((tab, index) => {
        id: group ~ '-' ~ tab.identifier,
        label: tab.name,
        content: tab.view,
        active: index == 0,
    })
```

GrabowskiM (changes requested):

lucasOsti (approved):

dew326 (approved):

GrabowskiM (approved):

PR #106, opened 2021-09-24
IBX-827 Moving Jobs on 1.0 ez-support-tools 

kacper-wieczorek-ibexa:


mnocon (changes requested):
Please add unit tests execution to Github Actions (only CS is executed right now).

After it's done we can get rid of the Travis configuration, because all of the jobs will be on Github Actions (only CS and Unit tests are on this branch).

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup PHP Action
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+          coverage: none
+          extensions: redis
```
mnocon, line 37: Why is this different than the standard value in the template (`extensions: pdo_sqlite, gd`),  what difference does it make?

mnocon (approved):

Steveb-p (changes requested):
YAML indentation is incorrect. It should be 4 spaces.

